PR #474, opened 2023-07-24
Demographic changes

### LINKED ISSUE #459: [Enhancement]: Demographics Questionnaire Changes

Currently, the demographics questionnaire is implemented in `/instruments`. This is not an ideal setup (I want to implement in pure JSON with type-safe schema, but I haven't figured out how to do that yet).

### Changes

Simon Ducharme, a doctor who is planning on using the platform in his clinic for dementia patients, has requested we add the following additional options:

- [x] Religion - add "none" and "agnostic"
- [x] Ethnic origin - add "mixed" (would be practical to be able to type letters to find best option rather than scroll)
- [x] Employment status - add "retired"

After you have changed the code, you will need to insert the updated instruments in the database. The easiest way you will be able to do this is to drop the database and run the setup again. 

### Estimated Difficulty

Low

### Priority

High

RyanHaniff:
Closes #459


joshunrau (approved):
very good!

Issue #365, opened 2023-05-08
[Feature]: Add Robots.txt


gdevenyi:
### Current Implementation

None

### Limitations of Current Implementation

Should have one

### Associated Application Components

Client

### Proposed Solution

Deny everything

### Estimated Difficulty

Low

### Priority

Low

joshunrau:
This should be tagged as feature if it does not exist

PR #476, opened 2023-07-25
added robots.txt to block all web crawlers from the whole site

Flyinchicken:


joshunrau (approved):
great!

joshunrau:
Also, @Flyinchicken for future reference you can close the issue with something like: Closes #365

PR #631, opened 2024-01-30
Add encryption and auth to gateway

joshunrau:


gdevenyi:
PR summary please :+1:

apps/api/src/configuration/configuration.schema.ts
```diff
@@ -3,6 +3,7 @@ import { z } from 'zod';
 
 export const $Configuration = z.object({
   API_DEV_SERVER_PORT: z.coerce.number().positive().int().optional(),
+  GATEWAY_API_KEY: z.string().min(32),
```
gdevenyi, line 6: If you require a minimum size, say so in the env template.

packages/database/prisma/core-schema.prisma
```diff
@@ -17,17 +17,25 @@ enum AssignmentStatus {
...
-  instrumentRecord InstrumentRecordModel?
-  status           AssignmentStatus
-  subject          SubjectModel           @relation(fields: [subjectId], references: [id])
-  subjectId        String
-  url              String
+  createdAt         DateTime               @default(now()) @db.Date
+  updatedAt         DateTime               @updatedAt @db.Date
+  id                String                 @id @map("_id")
+  completedAt       DateTime?
+  expiresAt         DateTime
```
gdevenyi, line 30: Optional expiry?

gdevenyi (changes requested):

joshunrau:
> PR summary please 👍

For future reference:


Okay, so how the gateway authentication/security works:
1.	The environment variable `GATEWAY_API_KEY` is defined as the same value for both the API behind the firewall (called the “Core API”) and the API outside the Douglas (called the “Gateway”).
2.	A user at the Douglas creates an assignment, which sends a request to the Core API
3.	The Core API generates a public-private key pair, saves it in the MongoDB database (called the "Core Database") within an `Assignment`.
4.	The Core API sends a request to the Gateway including the public key, the instrument bundle, and the assignment. The HTTP request header `Authorization` is set to `Bearer ${GATEWAY_API_KEY}`.
5.	The gateway accepts `GET` requests to `/assignments/:id`, `GET` and `POST` requests to `/api/assignments`, and `PATCH` and `DELETE` requests to `/api/assignments/:id`. For the API routes (i.e., all requests except for `GET /assignments/:id`), the authentication procedure is initiated. 
6. In the authentication procedure, the token is extracted from the `Authorization` header. First, it checks if the token matches the `GATEWAY_API_KEY`, in which case the request proceeds. Next, it checks whether the request is `PATCH /api/assignments/:id` AND the token is equal to the result of `generateToken(id)`, where the return value is the `sha256` hash of the assignment uuid appended to the `GATEWAY_API_KEY`. If this is the case, the request can proceed, otherwise it returns `401 Unauthorized`.
7. When the Core API sends `POST /api/assignments`, the Gateway therefore checks the `GATEWAY_API_KEY`, validates the data, and inserts it into the SQLite database (called the "Gateway Database").
7. When the user visits `/assignments/:id`, the Gateway checks in the Gateway Database whether or not an assignment with that ID exists. If not, then it returns 404. Otherwise, it dynamically generates the html page using [`react-dom/server`](https://react.dev/reference/react-dom/server), assigning the token to the global window object.
8. The instrument then renders to the user. When it is completed, the static page rendered by the Gateway sends `PATCH /api/assignments/:id` with the generated token in the headers, and the data from the instrument in the body.
9. The Gateway validates this data and encrypts it using the public key, after which point it is inserted into the Gateway Database. 
10. Periodically (default = every 10 seconds), the Core API sends `GET /api/assignments` to the Gateway. If there are any complete assignments, it then decrypts the data, creates an `InstrumentRecord` in the Core Database, updates the `Assignment` in the Core Database, and sends `DELETE /api/assignments/:id` to the Gateway.

apps/api/src/configuration/configuration.schema.ts
joshunrau: bb9e463fe44bcb7c9f09b3f1c0be9fa960d507d2

packages/database/prisma/core-schema.prisma
joshunrau: I added a default expiry of one year. Much easier to do on client side than change the internal workings everywhere for this. If you think this should be longer, let me know and I can change in next PR.

PR #634, opened 2024-01-31
Interactive summary

joshunrau:
Rewrites `InstrumentMeasures` to a discriminated union with kinds `const` and `computed`. Computed corresponds to the previous data structure, with the exception that the value may be a `boolean`, `number`, or `string` instead of only string. For the purposes of graphing, non-numeric values are ignored.

On the other hand, `const` includes a string `ref` which, for an arbitrary instrument with data type `T`, is `keyof T` if `T` is a plain object, otherwise `never` (i.e., `const` is not a valid option in such cases).

For instruments where the data is graphable, this new format reduces boilerplate significantly, 

Please see the implementation below:

```typescript
type InstrumentMeasures<
  TData = any,
  TLanguage extends InstrumentLanguage = InstrumentLanguage,
> = Record<
  string,
  | {
      kind: "computed";
      label: InstrumentUIOption<TLanguage, string>;
      value: (data: TData) => InstrumentMeasureValue;
    }
  | {
      kind: "const";
      label?: InstrumentUIOption<TLanguage, string>;
      ref: TData extends Record<string, any>
        ? ConditionalKeys<TData, InstrumentMeasureValue> extends infer K
          ? [K] extends [never]
            ? string
            : K
          : never
        : never;
    }
>;
```

Using this new data structure, measures replace the previous `InstrumentSummary` implementation, where `FormDataType` was the assumed format for the summary. The existing form instruments have been adjusted to account for this new format. 

For example, consider the MoCA:

**Before:**

```javascript
{
  measures: {
    abstraction: {
      label: {
        en: 'Abstraction',
        fr: 'Abstraction'
      },
      value: ({ abstraction }) => abstraction
    },
    attention: {
      label: {
        en: 'Attention',
        fr: 'Attention'
      },
      value: ({ attention }) => attention
    },
    delayedRecall: {
      label: {
        en: 'Delayed Recall',
        fr: 'Rappel'
      },
      value: ({ delayedRecall }) => delayedRecall
    },
    language: {
      label: {
        en: 'Language',
        fr: 'Langue'
      },
      value: ({ language }) => language
    },
    naming: {
      label: {
        en: 'Naming',
        fr: 'Dénomination'
      },
      value: ({ naming }) => naming
    },
    orientation: {
      label: {
        en: 'Orientation',
        fr: 'Orientation'
      },
      value: ({ orientation }) => orientation
    },
    totalScore: {
      label: {
        en: 'Total Score',
        fr: 'Score total'
      },
      value: (data) => {
        let sum = 0;
        Object.values(data).forEach((value) => {
          sum += Number(value);
        });
        return sum;
      }
    }
  }
}

```

**After:**

```javascript
{
  measures: {
    abstraction: {
      kind: 'const',
      ref: 'abstraction'
    },
    attention: {
      kind: 'const',
      ref: 'attention'
    },
    delayedRecall: {
      kind: 'const',
      ref: 'delayedRecall'
    },
    language: {
      kind: 'const',
      ref: 'language'
    },
    naming: {
      kind: 'const',
      ref: 'naming'
    },
    orientation: {
      kind: 'const',
      ref: 'orientation'
    },
    totalScore: {
      kind: 'computed',
      label: {
        en: 'Total Score',
        fr: 'Score total'
      },
      value: (data) => {
        let sum = 0;
        Object.values(data).forEach((value) => {
          sum += Number(value);
        });
        return sum;
      }
    }
  }
}

```


gdevenyi (approved):
Very cool, I like the simplification.

PR #722, opened 2024-02-27
FRA translation review

ValiquetteVanessa:
Reviewing the FRA translation of the Open Data Capture documentation available online.


joshunrau (approved):
merci

Issue #907, opened 2024-08-08
Do not produce duplicate form from link if files contents are the same.


david-roper:
### Current Implementation and Limitations

The current implementation produces a new form with and index number whenever a link is clicked an that form already exists in local cache ex Happiness_Questionnaire (1), Happiness_Questionnaire (2), etc.. This can lead to multiple duplicate forms if the link is used multiple times on the same device.

### Associated Application Components

Client

### Proposed Solution

Compare the file contents of the link form to the file contents of the already existing one. If they are the same the duplicate is not created. Start within the playground IndexPage.tsx and go from there (apps/playground/src/pages/IndexPage.tsx). 

### Estimated Difficulty

Low

### Priority

Low

PR #915, opened 2024-08-13
Check instrument duplicate

david-roper:
closes issue #907 

useEffect within playgrounds indexPage now checks if form from a link contains the same name and content of a previously existing form. If so it redirects the user the that previous form. Name+suffix convention is now reserved for a linked form with the same name but different content.


apps/playground/src/pages/IndexPage.tsx
joshunrau, line 1: There could be multiple files for an instrument. Please check that all file names and content is the same.

joshunrau (changes requested):

joshunrau (approved):

PR #920, opened 2024-08-18
major internal changes

joshunrau:


gdevenyi:
Can you provide some commentary on what this is.

A single commit is a bit hard to review.

joshunrau:
Notable Changes
- Create `vendor` folder, which contains wrappers and types for 3rd party dependencies
- Create new `runtime-bundler` package which generates the runtime based on a config file (see `runtime/v1/runtime.config.js`based on ESBuild (several milliseconds) and TypeScript compiler API (still very slow, but at least works)
- Removed reliance on`rollup-plugin-dts` which is not capable of handing the size of the runtime
- Move out packages `crypto` and `json-utils` into `libcrypto` and `libjs` respectively
- Resolve circular dependencies
- Create new `runtime-core` package which is relied upon by many things
- Use vendor package for `zod`

PR #967, opened 2024-10-04
replace site address with :80

thomasbeaudry:
you need :80 so you the system's caddy talk to your internal caddy

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -1,4 +1,4 @@
-{$SITE_ADDRESS} {
+:80 {
 	redir /api /api/
 
 	handle_path /api/* {


joshunrau (changes requested):
We need the environment variable for reverse proxy setups (like what I have for the demo). Please change the `SITE_ADDRESS` variable in `.env.template` instead of hardcoding it here.

Also, I was under the impression that `:80` is just a shorthand for`localhost:80`? Are there some docs I can read to learn about the distinction that you found?

gdevenyi:
:80 is not equivalent to localhost:80

The cicvmhost01 deployment was unable to function in a caddy-behind-caddy (system caddy + docker caddy) situation without making this change.

system caddy config was:
```
#Internal OpenDataCapture for human studies
http://odc.douglas.rtss.qc.ca {
        reverse_proxy http://localhost:5500
}
```

joshunrau:
Okay I'll change this in the env file. We need a generic solution for this

joshunrau:
> > Okay I'll change this in the env file. We need a generic solution for this
> 
> You mean that the SITE_ADDRESS will become the entire line in the caddyfile?

No, I mean I will change `localhost:80` to ` :80`.

joshunrau:
This is now changed in the template

Issue #1002, opened 2024-10-22
Set Example For Upload Template is Generated WIth Incorrect Values


joshunrau:
For the Happiness Questionnaire, the template example for `causesOfDissatisfaction` is given as `SET(a,b,c) (optional)`. In the schema, it is defined as ` z.set(z.enum(['MONEY', 'FRIENDS', 'ROMANTIC_PARTNER', 'EXISTENTIAL_CRISIS'])).optional()`

david-roper:
Yeah this was one of the previous TODOs, will work on it now

Issue #918, opened 2024-08-14
Import data session


david-roper:
### Description

Create a type of session in which the user can log an ID and import data via csv. The step by step process should go something like this:

- select custom identifier for session
- select import data session
- select csv file to import and which instrument to import the data into
- let instrument validation schema validate csv data
- if valid import data into the subject table

### Estimated Difficulty

High

### Priority

Medium

PR #1006, opened 2024-10-23
Set sample data

david-roper:
create sample data for set fields when a template is generated for example

z.set(z.enum([a,b,c])

becomes sample data SET(a/b/c, ...)

closes issue #1002 

related to issue #918

### DIFF
--- a/apps/web/src/features/upload/utils.ts
+++ b/apps/web/src/features/upload/utils.ts
@@ -73,6 +73,10 @@ function isZodEnumDef(def: AnyZodTypeDef): def is z.ZodEnumDef {
   return def.typeName === z.ZodFirstPartyTypeKind.ZodEnum;
 }
 
+function isZodSetDef(def: AnyZodTypeDef): def is z.ZodSetDef {
+  return def.typeName === z.ZodFirstPartyTypeKind.ZodSet;
+}
+
 function isZodArrayDef(def: AnyZodTypeDef): def is z.ZodArrayDef {
   return def.typeName === 'ZodArray';
 }
@@ -126,6 +130,7 @@ export function reformatInstrumentData({
 
 export function getZodTypeName(schema: z.ZodTypeAny, isOptional?: boolean): ZodTypeNameResult {
   const def: unknown = schema._def;
+  console.log('zod def', def);
   if (isZodTypeDef(def)) {
     if (isZodOptionalDef(def)) {
       return getZodTypeName(def.innerType, true);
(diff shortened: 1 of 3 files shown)


apps/web/src/features/upload/hooks/useProcessUploadData.ts
joshunrau, line 1: Why does this exist?

apps/web/src/features/upload/utils.ts
```diff
@@ -73,6 +73,10 @@ function isZodEnumDef(def: AnyZodTypeDef): def is z.ZodEnumDef {
   return def.typeName === z.ZodFirstPartyTypeKind.ZodEnum;
 }
 
+function isZodSetDef(def: AnyZodTypeDef): def is z.ZodSetDef {
+  return def.typeName === z.ZodFirstPartyTypeKind.ZodSet;
+}
+
 function isZodArrayDef(def: AnyZodTypeDef): def is z.ZodArrayDef {
   return def.typeName === 'ZodArray';
```
joshunrau, line 81: You should use `z.ZodFirstPartyTypeKind` in all cases to be consistent

apps/web/src/features/upload/utils.ts
```diff
@@ -138,7 +142,31 @@ export function getZodTypeName(schema: z.ZodTypeAny, isOptional?: boolean): ZodT
       };
     } else if (isZodArrayDef(def)) {
       return interpretZodArray(schema, def.typeName, isOptional);
+    } else if (isZodSetDef(def)) {
+      const innerDef = def.valueType._def as AnyZodTypeDef;
```
joshunrau, line 146: We should not introduce any new code like this. Please write a type safe function to validate and parse the `_def`. This can also fix other type issues that we have.

joshunrau: For this feature to no longer be marked as experimental, there should be no `as` statements, except with detailed comment expressing why this is necessary.

apps/web/src/features/upload/utils.ts
```diff
@@ -138,7 +142,31 @@ export function getZodTypeName(schema: z.ZodTypeAny, isOptional?: boolean): ZodT
...
+      if (!isZodTypeDef(innerDef)) {
+        return {
+          message: 'Invalid inner type: ZodSet value type must have a valid type definition',
+          success: false
+        };
+      }
+
+      if (!isZodEnumDef(innerDef)) {
+        return {
+          message: 'Invalid inner type: ZodSet value type must be a ZodEnum',
```
joshunrau, line 157: Why?

apps/web/src/features/upload/utils.ts
```diff
@@ -315,7 +343,17 @@ function generateSampleData({
     case 'ZodNumber':
       return formatTypeInfo('number', isOptional);
     case 'ZodSet':
-      return formatTypeInfo('SET(a,b,c)', isOptional);
+      try {
+        if (!enumValues) {
```
joshunrau, line 347: it is not necessarily an enum

joshunrau (changes requested):
This is not a proper implementation. It incorrectly assumes that all Set fields are enums.

PR #1075, opened 2025-01-14
add support for importing mp4 videos in instruments

joshunrau:
Please see example for usage. 

Also adds error message for upload failure and allows uploading zipped directory


gdevenyi:
Where is the data going? In the database?

Performance will be abysmal 

This is what the S3 backend stuff is for

Issue #1118, opened 2025-04-07
Improve performance of upload csv feature (threading, chunks, etc.)


david-roper:
### Current Implementation and Limitations

Look into way to improve the upload features performance. Allow for users to traverse the website whilst upload is still going.

### Associated Application Components

Client

### Proposed Solution

Looking into the possibility of threading the process. Turn the database entries from the request into chunks that are done by the threads. 

### Estimated Difficulty

None

### Priority

High

PR #1122, opened 2025-04-14
Refactor creation of subject ids in upload csv code.

david-roper:
Works on issue #1118 

changes: 

Removed createSubjectIfNotFound function 
Replaced with a createMany function within subjectService which creates all the subject ids needed for an upload in one go.

### DIFF
--- a/apps/api/src/instrument-records/instrument-records.service.ts
+++ b/apps/api/src/instrument-records/instrument-records.service.ts
@@ -15,6 +15,7 @@ import type {
 import { Prisma } from '@prisma/client';
 import type { Session } from '@prisma/client';
 import { isNumber, pickBy } from 'lodash-es';
+import pLimit from 'p-limit';
 
 import type { EntityOperationOptions } from '@/core/types';
 import { GroupsService } from '@/groups/groups.service';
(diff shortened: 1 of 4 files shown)


apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -276,6 +276,16 @@ export class InstrumentRecordsService {
     const createdSessionsArray: Session[] = [];
 
     try {
+      const subjectIdList = records.map((record) => {
```
joshunrau, line 279: const subjectIdList = records.map(({ subjectId }) => {

apps/api/src/subjects/subjects.service.ts
```diff
@@ -41,6 +41,43 @@ export class SubjectsService {
     });
   }
 
+  async createMany(data: CreateSubjectDto[]) {
+    //filter out all duplicate ids that are planned to be created
+
+    const seen = new Set();
```
joshunrau, line 47: why are you using a Set here?

apps/api/src/subjects/subjects.service.ts
```diff
@@ -41,6 +41,43 @@ export class SubjectsService {
...
+    const noDuplicateSubjectsList = data.filter((record) => {
+      if (!seen.has(record.id)) {
+        seen.add(record.id);
+        return true;
+      }
+      return false;
+    });
+
+    const subjectIds = noDuplicateSubjectsList.map((record) => record.id);
```
joshunrau, line 56: you could just create a set from the list of all subject ids also

apps/api/src/subjects/subjects.service.ts
```diff
@@ -41,6 +41,43 @@ export class SubjectsService {
...
+    //find the list of subject ids that already exist
+    const existingSubjects = await this.subjectModel.findMany({
+      select: { id: true },
+      where: {
+        id: { in: subjectIds }
+      }
+    });
+
+    //create a set of existing ids in the database to filter our to-be-created ids with
+    const existingIds = new Set(existingSubjects.map((subj) => subj.id));
```
joshunrau, line 67: won't this be equivalent to subjectIds? maybe I'm missing something?

joshunrau (changes requested):

apps/api/src/subjects/subjects.service.ts
david-roper: I use set in this case for the .has() method, makes it easier to find out if im adding a possible duplicate as it makes is just one boolean statement.

apps/api/src/subjects/subjects.service.ts
david-roper: existing ids includes only the ids already in the database. SubjectIds contains all the ones we'd like to be created without any duplicates. I compare both so have a final list of ids that are unique and do not already exist in the database.

gdevenyi:
Did you benchmark this?

Issue #1139, opened 2025-05-16
Expanding zod array data into entries with datahub export


david-roper:
### Current Implementation and Limitations

Current behaviour of the datahub export has the zod array data all in one entry in its json object like form. It looks out of place compared to the other long data format entries

### Associated Application Components

Client

### Proposed Solution

when zod array like data is found within the entry. iterate through each part and make it a standalone entry in the datahub's long format data export

### Estimated Difficulty

Low

### Priority

Medium

david-roper:
If we go with the former (ODC itself), we could detect it in the type of data sent over and expand it when its a zod array object. Would have to look at other options in the case of the wrapper doing it.

gdevenyi:
To make it clear, my preference is that the author of the instrument should fully define the data it returns in a way that its useful downstream, i.e. the unpacking happens in the instrument definition. The author of this should know best compared to a "generic" ODC unpacker.

PR #1146, opened 2025-05-26
Expand array export

david-roper:
works on issue #1139

Takes in data, if its a zod array the data is expanded into separate csv entries


gdevenyi:
Can we have some example before/after exports in the PR message?

david-roper:
I will have to retest this branch to see if it can be improved with the new zod4 changes

david-roper:
Initial input 
image

CSV output

outputTestRecordArray

packages/runtime-core/src/types/instrument.base.ts
```diff
@@ -104,7 +105,7 @@ type UnilingualClientInstrumentDetails = ClientInstrumentDetails<Language>;
 type MultilingualClientInstrumentDetails = ClientInstrumentDetails<Language[]>;
 
 /** @public */
-type InstrumentMeasureValue = boolean | Date | number | string | undefined;
+type InstrumentMeasureValue = boolean | Date | number | RecordArrayFieldValue[] | string | undefined;
```
joshunrau, line 108: define a schema for `$RecordArrayFieldValue` in `packages/schemas`

packages/schemas/src/instrument/instrument.base.ts
```diff
@@ -97,6 +97,7 @@ const $InstrumentMeasureValue: z.ZodType<InstrumentMeasureValue> = z.union([
   z.boolean(),
   z.number(),
   z.date(),
+  z.array(z.any()),
```
joshunrau: `$RecordArrayFieldValue`

joshunrau: Infer the type and export it from schemas

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -378,6 +404,36 @@ export class InstrumentRecordsService {
     }
   }
 
+  private expandData(
```
joshunrau: should be able to type this as `$RecordArrayFieldValue`

david-roper: Since the $RecordArrayFieldValue is a zod schema i added a safeparse check of the value instead

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -153,26 +166,40 @@ export class InstrumentRecordsService {
...
+            sessionDate: record.session.date.toISOString(),
+            sessionId: record.session.id,
+            sessionType: record.session.type,
+            subjectAge: record.subject.dateOfBirth ? yearsPassed(record.subject.dateOfBirth) : null,
+            subjectId: record.subject.id,
+            subjectSex: record.subject.sex,
+            timestamp: record.date.toISOString()
+          };
+          const arrayResult = this.expandData(data, measureValue, instrument, objectRecord);
+          if (arrayResult.isErr()) {
```
joshunrau: we need to give the detailed error. Just throw in your expandData, or log the error.

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -153,26 +166,40 @@ export class InstrumentRecordsService {
...
+            groupId: record.subject.groupIds[0] ?? DEFAULT_GROUP_NAME,
+            sessionDate: record.session.date.toISOString(),
+            sessionId: record.session.id,
+            sessionType: record.session.type,
+            subjectAge: record.subject.dateOfBirth ? yearsPassed(record.subject.dateOfBirth) : null,
+            subjectId: record.subject.id,
+            subjectSex: record.subject.sex,
+            timestamp: record.date.toISOString()
+          };
+          const arrayResult = this.expandData(data, measureValue, instrument, objectRecord);
```
joshunrau: what exactly are you doing here? I assume you are mutating the array? Why not a pure function?

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -378,6 +405,38 @@ export class InstrumentRecordsService {
...
+    data: InstrumentRecordsExport,
+    listEntry: any[],
+    instrument: ScalarInstrument,
+    record: RecordObject
+  ) {
+    if (listEntry.length > 0) {
+      for (const objectEntry of listEntry) {
+        for (const [dataKey, dataValue] of Object.entries(objectEntry as { [key: string]: any })) {
+          const parseResult = $RecordArrayFieldValue.safeParse(dataValue);
+          if (!parseResult.success) return err('Invalid data value');
```
joshunrau: how are we supposed to debug with this error message if there is a problem?

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -378,6 +405,38 @@ export class InstrumentRecordsService {
...
+            instrumentName: instrument.internal.name,
+            measure: dataKey,
+            sessionDate: record.sessionDate,
+            sessionId: record.sessionId,
+            sessionType: record.sessionType,
+            subjectAge: record.subjectAge,
+            subjectId: record.subjectId,
+            subjectSex: record.subjectSex,
+            timestamp: record.timestamp,
+            value: typeof dataValue === 'string' ? dataValue : JSON.stringify(dataValue)
```
joshunrau: why does the value need to be stringified? it should already be stringified when you return it via json encoding

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -378,6 +405,38 @@ export class InstrumentRecordsService {
...
+          });
+        }
+      }
+      return ok('Success');
+    } else {
```
joshunrau: https://www.youtube.com/watch?v=CFRhGnuXG-4

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -378,6 +405,38 @@ export class InstrumentRecordsService {
...
+      return err('Error interpreting array');
```
joshunrau: how are we supposed to debug with this error message if there is a problem?

joshunrau (changes requested):

joshunrau (approved):

Issue #1168, opened 2025-07-16
Add the ability to delete/update subjects in ODC-CLI


david-roper:
### Current Implementation and Limitations

Currently only instrument records can be update and deleted. It would be beneficial to have this feature work for subject too to deal with test data/ typos

### Associated Application Components

Client, Server

### Proposed Solution

Add an update and delete option for the subject service backend
Add a subject section with find, delete, and update commands in the odc cli

### Estimated Difficulty

Medium

### Priority

Medium

PR #1170, opened 2025-07-21
Subject cli commands

david-roper:
Works on issue #1168 

Having issues with calling instrumentsService in the subjectService delete instrument records attached to a subject id.

[error.json](https://github.com/user-attachments/files/21353343/error.json)

This might be due to a circular dependency in the subject and instrument record modules relying on eachother.

Apart from that the find commands works well.

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands to find and delete subjects, including confirmation prompts for deletions and filtering options when searching.
* **Improvements**
  * Enhanced API input validation for subject deletion, ensuring proper handling of encoded subject IDs.

### DIFF
--- a/apps/api/src/instrument-records/instrument-records.module.ts
+++ b/apps/api/src/instrument-records/instrument-records.module.ts
@@ -1,4 +1,4 @@
-import { Module } from '@nestjs/common';
+import { forwardRef, Module } from '@nestjs/common';
 
 import { GroupsModule } from '@/groups/groups.module';
 import { InstrumentsModule } from '@/instruments/instruments.module';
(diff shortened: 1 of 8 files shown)


apps/api/src/instrument-records/instrument-records.module.ts
```diff
@@ -12,7 +12,7 @@ import { InstrumentRecordsService } from './instrument-records.service';
 @Module({
   controllers: [InstrumentRecordsController],
```
joshunrau: Remove

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -34,7 +41,8 @@ export class InstrumentRecordsService {
     private readonly instrumentMeasuresService: InstrumentMeasuresService,
     private readonly instrumentsService: InstrumentsService,
     private readonly sessionsService: SessionsService,
-    private readonly subjectsService: SubjectsService
+    @Inject(forwardRef(() => SubjectsService))
+    private readonly subjectsService: InstanceType<typeof SubjectsService>
```
joshunrau: remove

PR #1238, opened 2025-11-21
Refactors of useFindSessionQuery, useEffect, and nullable subject in $Session schema

david-roper:
Add the changes from joshs changes and fixes typing issues from adding nullable() to subject in the $Session schema

### DIFF
--- a/apps/web/src/hooks/useFindSessionQuery.ts
+++ b/apps/web/src/hooks/useFindSessionQuery.ts
@@ -26,12 +26,7 @@ export const useFindSessionQuery = (
       const response = await axios.get('/v1/sessions', {
         params
       });
-      const parsedData = $SessionWithUser.array().safeParseAsync(response.data);
-      if ((await parsedData).error) {
-        const message = (await parsedData).error?.message;
-        throw new Error(message);
-      }
-      return (await parsedData).data;
+      return $SessionWithUser.array().parseAsync(response.data);
     },
     queryKey: ['sessions', ...Object.values(params)]
   });


apps/web/src/hooks/useFindSessionQuery.ts
```diff
@@ -36,10 +23,14 @@ export const useFindSessionQuery = (
   return useQuery({
     enabled,
     queryFn: async () => {
-      const response = await axios.get('/v1/sessions/', {
+      const response = await axios.get('/v1/sessions', {
         params
       });
-      return $SessionWithUser.array().parseAsync(response.data);
+      const parsedData = $SessionWithUser.array().safeParseAsync(response.data);
```
joshunrau: how is this an improvement compared to the Zod error? This error message is terrible.

apps/web/src/routes/_app/instruments/render/$id.tsx
```diff
@@ -61,7 +61,7 @@ const RouteComponent = () => {
       <div className="grow">
         <InstrumentRenderer
           className="mx-auto max-w-3xl"
-          subject={currentSession?.subject}
+          subject={currentSession?.subject ?? undefined}
```
joshunrau, line 64: this is confusing. just change the subject prop in instrument renderer to accept null or undefined

apps/web/src/hooks/useFindSessionQuery.ts
david-roper: true changing it now

apps/web/src/routes/_app/instruments/render/$id.tsx
david-roper: if i update instrument renderer i will have to update  3 different instrument types to accept null (series, scalar instruments) with basically the same solution. Would it better to change this file or all three props of the instrument display related components `ScalarInstrumentRenderer ` `SeriesInstrumentRenderer` `InstrumentSummary`?

joshunrau: okay fair enough, we can stick with this

apps/web/src/hooks/useFindSessionQuery.ts
joshunrau: this is less bad, but I don't understand the point of throwing a different error, instead of just throwing the zod error like I had before? Maybe I am missing something? This approach just means you will have fewer details than the full zod error.

Issue #1220, opened 2025-10-23
For all instruments, log ODC user who initiated the session and export as part of the data


gdevenyi:
### Current Implementation and Limitations

Data audit log/provenience is not complete without user information

### Associated Application Components

Server, Client

### Proposed Solution

Session should contain ODC user information, which is already linked to instrument records. Exports should include this information

### Estimated Difficulty

None

### Priority

Medium

PR #1244, opened 2025-11-26
Add username to export

david-roper:
closes issue #1220 

## Summary by CodeRabbit

* **New Features**
  * List sessions with associated usernames via a new API and client hook.
  * Sessions now support an augmented session-with-user shape and optional subject.
  * Export outputs (CSV, TSV, Excel, JSON) include a Username column in wide and long formats.

* **Bug Fixes**
  * Prevents errors when session subjects or linked groups are missing.
  * Removes duplicate sheet creation in Excel exports.

* **Tests**
  * Updated tests to validate username inclusion in export outputs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


apps/api/src/sessions/sessions.service.ts
```diff
@@ -94,6 +96,26 @@ export class SessionsService {
...
+          select: {
+            username: true
+          }
+        }
+      },
+      where: {
+        AND: [accessibleQuery(ability, 'read', 'Session'), { groupId }]
+      }
+    });
+    if (!sessionsWithUsers) {
```
joshunrau: this will never be true, as empty array is truthy. you should check .length

apps/web/src/hooks/useInstrumentVisualization.ts
```diff
@@ -195,19 +215,42 @@ export function useInstrumentVisualization({ params }: UseInstrumentVisualizatio
...
+            en: 'Error occurred finding records',
+            fr: "Une erreur s'est produite lors de la recherche des enregistrements."
+          }),
+          type: 'error'
         });
       }
-      setRecords(records);
-    }
+    };
+    void fetchRecords();
```
joshunrau: again, why is this a function?

apps/web/src/hooks/useInstrumentVisualization.ts
```diff
@@ -195,19 +215,42 @@ export function useInstrumentVisualization({ params }: UseInstrumentVisualizatio
...
+              username: username,
+              ...record.computedMeasures,
+              ...props
+            };
+          });
+
+          setRecords(records);
+        }
+      } catch (error) {
+        console.error('Error occurred: ', error);
```
joshunrau: just log the error. it is implied an error occurred if an error is logged.

joshunrau (changes requested):

apps/web/src/hooks/useInstrumentVisualization.ts
david-roper: This can be removed since initially it was a asynchronous function to use the previous api calls. good catch.

apps/web/src/hooks/useInstrumentVisualization.ts
david-roper: using console.log is currently not allowed by the linter

joshunrau (approved):

Issue #1247, opened 2025-11-28
Order of DataHub tabs should be different


gdevenyi:
### Current Implementation and Limitations

Currently the order starts with assignments

### Associated Application Components

Client

### Proposed Solution

If I am looking up a subject's data I don't expect the first thing to be the assignments.

### Estimated Difficulty

Low

### Priority

Low

Issue #1230, opened 2025-11-13
Improve datahub search by providing list of ids matching the search string


david-roper:
### Current Implementation and Limitations

currently the search bar only matches a subject when there is a one to one match with the id and the search string However with some ids being quite long this can be quite clunky to use. Furthermore, when a search is successful the page instantly traverses user to that subjects table page.

### Associated Application Components

Client

### Proposed Solution

After discussing with the mouseODC users it may be more useful to have the search bar match with multiple ids and to display those subjects that matched in the datahub table. which allows users to more easily find the subjects they created and click on them independently if they would like to view their subject tables. 

### Estimated Difficulty

Medium

### Priority

High

PR #1252, opened 2025-12-05
Improve datahub search

david-roper:
Working on issue #1230 
Also works on issue #1247 

- Adds a new search bar which matches ids to the search string
- matched ids are displayed in datahub table
- export feature now can export ids that matched the search
- ids and lookup now navigates to subject table page instead of assignment page
- Added search bar and its info in the walkthrough
- made the subject lookup as separate button.

## Summary by CodeRabbit

* **New Features**
  * Header search bar for real-time subject filtering in the Datahub table.
  * Subject lookup opened via a dedicated search button and dialog.

* **Improvements**
  * Exports (CSV, Excel, JSON) now export only currently filtered subjects, require visible selections, and guard against empty exports.
  * Export notifications are shown after filtering and include more specific error messages.
  * Subject lookup now navigates directly to the subject table view.

* **Documentation**
  * Walkthrough updated with an added step and localized guidance for the header search and lookup button.

* **Style**
  * Spotlight selector updated to target the renamed subject search bar.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/web/src/providers/WalkthroughProvider.tsx
+++ b/apps/web/src/providers/WalkthroughProvider.tsx
@@ -160,7 +160,7 @@ const Walkthrough = () => {
           <p>
             {t({
               en: 'Here, you can search for subjects in the database. To begin, click on the Subject Lookup button, and a popup will appear where you can enter the search query.',
-              fr: "Ici, vous pouvez rechercher des clients dans la base de données. Pour commencer, cliquez sur la button de recherche et une fenêtre contextuelle s'affichera pour vous permettre de saisir la requête de recherche."
+              fr: "Ici, vous pouvez rechercher des clients dans la base de données. Pour commencer, cliquez sur la bouton de recherche et une fenêtre contextuelle s'affichera pour vous permettre de saisir la requête de recherche."
             })}
           </p>
         ),


apps/web/src/providers/WalkthroughProvider.tsx
```diff
@@ -140,16 +140,35 @@ const Walkthrough = () => {
         content: (
           <p>
             {t({
-              en: 'Here, you can search for subjects in the database. To begin, click on the search bar, and a popup will appear where you can enter the search query.',
-              fr: "Ici, vous pouvez rechercher des clients dans la base de données. Pour commencer, cliquez sur la barre de recherche et une fenêtre contextuelle s'affichera pour vous permettre de saisir la requête de recherche."
+              en: 'Here, you can search for subjects in the database. To begin, type in the search bar to find matching subject identification tags',
```
joshunrau: can we clarify what these tags are by adding something like "(e.g., ID)", or just saying ID if it is only by ID

apps/web/src/routes/_app/datahub/index.tsx
```diff
@@ -83,24 +92,37 @@ const RouteComponent = () => {
...
     getExportRecords()
       .then((data): any => {
+        const listedSubjects = tableData.map((record) => {
+          return removeSubjectIdScope(record.id);
+        });
+
+        const filteredData = data.filter((dataEntry) => listedSubjects.includes(dataEntry.subjectId));
+
+        if (filteredData.length < 1) {
+          throw Error(
```
joshunrau, line 104: let's use a notification here instead of throwing an error so the user can see

apps/web/src/routes/_app/datahub/index.tsx
```diff
@@ -127,10 +156,25 @@ const RouteComponent = () => {
       setIsLookupOpen(false);
     } else {
       addNotification({ type: 'success' });
-      await navigate({ to: `./${response.data.id}/assignments` });
+      await navigate({ to: `./${response.data.id}/table` });
     }
   };
 
+  useEffect(() => {
```
joshunrau, line 163: this is good for now. once it is merged I will show you how we can improve in-person.

joshunrau (changes requested):

apps/web/src/routes/_app/datahub/index.tsx
david-roper: The error is thrown in order to be catched and used in the notification  at the end of the function. If the throw is not used the download of the empty file goes through

Issue #1254, opened 2025-12-11
add threading to datahub's export


thomasbeaudry:
### Current Implementation and Limitations

When you click on export in the datahub it starts to assemble your file in the current window (without giving you any UI feedback).  If you leave the page and navigate elsewhere in ODC t the export dies.  Exporting can take several minutes so the current implement isn't sufficient.

### Associated Application Components

Client, Server

### Proposed Solution

You need to use threading for downloads.  At the very least, I would use at least 1 thread that manages export.  You can also use several threads to assemble very large datasets if you want and create a bunch of temporary files and then merge them at the end if you want to speed things up (but you probably would want to have some logic to decide how many threads you need according to what is being asked for / how many records exist).

You also need to add some visual feedback to the user that the exporting process has begun and that a download popup will appear when the data is ready.

### Estimated Difficulty

Medium

### Priority

High

PR #1257, opened 2026-01-13
Add threading

david-roper:
Works on issue #1254 #1118 

Adds threading for the record processing.

However the true culprit of the poor performance is the initial prisma database call, which will have to be adjusted.


gdevenyi:
How does the benchmarking of this look? Or a flamegraph difference?

PR #1322, opened 2026-03-27
Add the $PROJECT_NAME variable to the .env file to customize the docker container'…

thomasbeaudry:
When setting up an ODC, it's easy to forgot to modify the docker-compose.yaml (and only focus on the .env), so by having this variable in the .env, you can completely ignore the docker-compose.yaml.  When managing multiple ODCs, this variable needs to be different for each instance, so it's useful to have

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -2,6 +2,9 @@
 ## PRODUCTION
 ## ---------------------------------
 
+# Project Name that will be prefixed to docker containers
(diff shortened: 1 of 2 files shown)


gdevenyi:
Does this show up anywhere on the front-end?

joshunrau (approved):

Issue #1342, opened 2026-04-28
Add username of uploader to data


david-roper:
### Content

Currently the username is attached to the data only when its collected via forms. Should this behavior be replicated in the upload feature? If so, do we assume the user logged in and uploading data is the one that collected the data as well?

gdevenyi:
Yes, the uploader should be the one associated with the data, there isn't really another choice

david-roper:
@gdevenyi  

@joshunrau made the good point that the uploader isn't necessarily the one who collected the data. He suggested that we could implement a checkbox to provide the option to attach the username onto the data or not.  What do you think?

gdevenyi:
If the user is bulk uploading, it is likely that the original person doesn't exist in ODC at all. I don't think this is worth it. If you upload, its your data now.

joshunrau:
> If the user is bulk uploading, it is likely that the original person doesn't exist in ODC at all. I don't think this is worth it. If you upload, its your data now.

In general, I agree. However, I think that it is flawed to assume this without any user intervention. In all other cases, the person recorded is known to be the person who collected the data; the person who uploaded the data is not the same. If you are using this for, as an example, determining effect of sex on the mice, then assuming this incorrectly will confound the data. In my opinion, it is much better to be explicit and require the user to sign off on this. Otherwise, if it is unknown who collected the data (or someone who does not exist in the system), better to have it null rather than have bad data. If people want to record free form text for who collected the data, they can add a form field for that.

gdevenyi:
How do we do this without polluting the data model. I do not want free-form text for "who collected it", as that will break the existing intention of "auto tracking the user as metadata"

david-roper:
@gdevenyi  @joshunrau I have now made a draft to pr to include this feature #1344, we can discuss whether to include the checkbox (i.e. make attaching username optional or not) before making the pr official.

PR #1350, opened 2026-05-05
Attach username to upload dropdown

david-roper:
The other option to close issue #1342 

This  pr removes the optional checkbox and make it always attach the user to the uploaded data whenever possible. 

If we select this option the pr #1344 will be removed

The new implementation now includes a dropdown of all the usernames in the group, which allows the user to select which username to attach to the upload. The dropdown defaults to the current user's username and also provides an 'N/A' option if needed. 

small change to update user backend that fixes issue #1354 

## Summary by CodeRabbit

* **New Features**
  * Option to select which username to associate with imported instrument records during upload.

* **Improvements**
  * User list is scoped to the active group so the dropdown shows relevant members.
  * Upload payload now includes the chosen username and the client forwards the selection.
  * Upload UI refined with a cleaner centered layout and loading presentation.

* **Bug Fixes**
  * Upload enforces that a chosen username belongs to the selected group.

* **Tests**
  * Added tests covering username handling during upload.

### DIFF
--- a/apps/api/src/instrument-records/instrument-records.service.ts
+++ b/apps/api/src/instrument-records/instrument-records.service.ts
@@ -347,7 +347,12 @@ export class InstrumentRecordsService {
     }
 
     if (username) {
-      await this.usersService.findByUsername(username, options);
+      const user = await this.usersService.findByUsername(username, options);
+      if (groupId && !user.groups.some((g) => g.id === groupId)) {
+        throw new ForbiddenException(
+          `User '${username}' is not a member of group '${groupId}'`
+        );
+      }
     }
 
     const createdSessionsArray: Session[] = [];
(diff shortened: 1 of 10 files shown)


apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -344,6 +352,13 @@ export class InstrumentRecordsService {
       );
     }
 
+    if (username) {
+      const user = await this.usersService.findByUsername(username, options);
+      if (groupId && !user.groups.some((g) => g.id === groupId)) {
```
joshunrau, line 357: if they are not admin, they should not be able to upload with no group id either

joshunrau (commented):
looks fine, let me know when ready to merge

joshunrau (approved):

PR #1363, opened 2026-06-01
Add --all mode to serve-instrument package

gdevenyi:
## Summary

Merges the multi-instrument serving capability directly into `serve-instrument` via a new `--all` CLI flag.

### Usage

**Single instrument (default, unchanged):**
```
serve-instrument ./my-instrument
```

**All instruments (new `--all` mode):**
```
serve-instrument --all ./instruments-dir
```

With `--all`, the target directory must contain `forms/` and/or `interactive/` subdirectories. The server provides:
- An **index page** linking to all discovered instruments
- **Per-instrument pages** with hot-rebuild on file change (`fs.watch`)
- An **en/fr language switcher** wired to the libui i18n translator
- **Verbose request logging** (with `-v`)

### Changes
- `src/cli.ts`: Added `--all` / `-a` flag; validates that target has `forms/`/`interactive/` subdirs in all mode
- `src/root.tsx`: Merged UI — `RootProps` is now a discriminated union (`single` | `index` | `instrument` pages); added `LanguageSwitcher` and `IndexPage` components
- `src/server.tsx`: Added `InstrumentLoaderMap` class and `AllModeHandler` alongside existing `SingleModeHandler`; `Server.create` branches on `mode`
- `package.json`: Bumped version to `0.0.13`

### DIFF
--- a/packages/serve-instrument/package.json
+++ b/packages/serve-instrument/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@opendatacapture/serve-instrument",
   "type": "module",
-  "version": "0.0.12",
+  "version": "0.0.13",
   "license": "Apache-2.0",
   "bin": {
     "serve-instrument": "./dist/cli.js"
--- a/packages/serve-instrument/src/cli.ts
+++ b/packages/serve-instrument/src/cli.ts
@@ -6,13 +6,20 @@ import { Command, InvalidArgumentError } from 'commander';
 import { name, version } from '../package.json';
 import { Server } from './server';
 
-function parseTarget(target: string): string {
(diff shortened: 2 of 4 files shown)


joshunrau (changes requested):
We should not be implementing a completely new package for this, with all of the code from serve-instrument copied near verbatim. You should use a subcommand instead so we can reuse the code and it can be maintainable in a single package.

gdevenyi:
> We should not be implementing a completely new package for this, with all of the code from serve-instrument copied near verbatim. You should use a subcommand instead so we can reuse the code and it can be maintainable in a single package.

Merged into serve-instrument package instead.

PR #1383, opened 2026-06-16
feat(instrument-repos): add GitHub-hosted instrument repositories

thomasbeaudry:
Admins can import instruments from a GitHub repository, sync to pull in new instruments, view a repo's instrument list, and delete repos. Group managers assign repos to groups and opt-in to individual instruments.

Backend:
- InstrumentRepo model + module/service/controller; instrument provenance tracked via Instrument.sourceRepoId/sourceRepoName
- private-repo support via an AES-256-GCM encrypted access token
- re-adding an existing repo re-syncs it (and updates a supplied token) instead of failing; repo name is no longer unique (url is the identity)
- orphan reconciliation when a source repo is deleted
- groups only auto-connect non-repo instruments on creation

Frontend:
- admin instrument-repos page (add/sync/view/delete) and repo assignment by the admin in manage groups; 
- Grouo manager can select instruments with insturment repo buttons displayed in group management
- Instrument summary & -preview feature added so group manager doesn't have to guess which form is which
- repo mutations invalidate the instrument list; opt-out flag added to the axios error interceptor so callers can show a single specific error toast

Tests: unit coverage for the repo service (crypto round-trip, secret stripping, delete guards, create/sync/dedup, orphan reconciliation).

## Release Notes

* **New Features**
  * Added instrument repository management: create/import, sync, list, view, and delete (with optional encrypted access tokens).
  * Admin “Instrument Repositories” page with add/sync/view/delete actions and safety checks when repos are assigned to groups.
  * Groups can link/unlink repositories, updating accessible instruments and exposing provenance for repo-sourced instruments.

* **Bug Fixes**
  * Improved group update logic for repository relations and name-collision handling.
  * Added ability/permission coverage for instrument repository management.

* **Tests**
  * Added unit tests covering encryption, import/sync/reconciliation, and deletion constraints.

* **Chores**
  * Updated CI ordering and Playwright version; improved request error-notification control.

### DIFF
--- a/apps/api/src/auth/ability.factory.ts
+++ b/apps/api/src/auth/ability.factory.ts
@@ -27,7 +27,7 @@ export class AbilityFactory {
         ability.can('manage', 'Assignment', { groupId: { in: groupIds } });
         ability.can('manage', 'Group', { id: { in: groupIds } });
         ability.can('read', 'Instrument');
-        ability.can('read', 'InstrumentRepo');
+        ability.can('read', 'InstrumentRepo', { groupIds: { hasSome: groupIds } });
         ability.can('create', 'InstrumentRecord');
         ability.can('create', 'InstrumentRecordFile', { groupId: { in: groupIds } });
         ability.can('read', 'InstrumentRecord', { groupId: { in: groupIds } });
--- a/apps/api/src/auth/ability.factory.ts
+++ b/apps/api/src/auth/ability.factory.ts
@@ -27,6 +27,7 @@ export class AbilityFactory {
         ability.can('manage', 'Assignment', { groupId: { in: groupIds } });
         ability.can('manage', 'Group', { id: { in: groupIds } });
         ability.can('read', 'Instrument');
+        ability.can('read', 'InstrumentRepo');
         ability.can('create', 'InstrumentRecord');
         ability.can('create', 'InstrumentRecordFile', { groupId: { in: groupIds } });
         ability.can('read', 'InstrumentRecord', { groupId: { in: groupIds } });
--- a/apps/api/src/instrument-repos/instrument-repos.service.ts
+++ b/apps/api/src/instrument-repos/instrument-repos.service.ts
@@ -5,7 +5,13 @@ import * as path from 'node:path';
 
 import { ConfigService, InjectModel, LoggingService } from '@douglasneuroinformatics/libnest';
 import type { Model } from '@douglasneuroinformatics/libnest';
-import { BadGatewayException, ConflictException, Injectable, NotFoundException } from '@nestjs/common';
+import {
+  BadGatewayException,
+  BadRequestException,
+  ConflictException,
+  Injectable,
+  NotFoundException
+} from '@nestjs/common';
 import type { OnModuleInit } from '@nestjs/common';
 import { bundle, BUNDLER_FILE_EXT_REGEX, inferLoader } from '@opendatacapture/instrument-bundler';
 import type { BundlerInput } from '@opendatacapture/instrument-bundler';
@@ -205,7 +211,9 @@ export class InstrumentReposService implements OnModuleInit {
       // https://github.com/.../archive/...: the API authenticates the request and then 302-redirects
       // to a *signed* codeload.github.com URL. fetch drops the Authorization header on that
       // cross-origin redirect, but the signed URL no longer needs it — so private repos still work.
-      const zipballUrl = `https://api.github.com/repos/${owner}/${repoName}/zipball/${branch}`;
+      // owner/repoName are already restricted to a strict allowlist in parseGitHubUrl; encode them
+      // anyway so nothing can break out of the intended api.github.com path component.
+      const zipballUrl = `https://api.github.com/repos/${encodeURIComponent(owner)}/${encodeURIComponent(repoName)}/zipball/${branch}`;
       try {
         const headers: { [key: string]: string } = {
           Accept: 'application/vnd.github+json',
@@ -363,8 +371,16 @@ export class InstrumentReposService implements OnModuleInit {
   private parseGitHubUrl(url: string): { owner: string; repoName: string } {
     const cleaned = this.normalizeUrl(url);
     const parts = cleaned.split('/');
-    const repoName = parts.pop()!;
-    const owner = parts.pop()!;
+    const repoName = parts.pop();
+    const owner = parts.pop();
+    // parts now holds ['https:', '', '<host>']; require the GitHub host and a strict owner/repo so a
+    // crafted URL cannot redirect the later GitHub API request elsewhere (server-side request forgery).
+    // GitHub owners/repos only ever contain these characters, so this rejects `@`, `:`, `?`, `#`, etc.
+    const host = parts[2];
+    const validName = /^[A-Za-z0-9._-]+$/;
+    if (host !== 'github.com' || !owner || !repoName || !validName.test(owner) || !validName.test(repoName)) {
+      throw new BadRequestException('Must be a valid GitHub repository URL (https://github.com/<owner>/<repo>).');
+    }
     return { owner, repoName };
   }
 
--- a/apps/api/src/instrument-repos/instrument-repos.service.ts
+++ b/apps/api/src/instrument-repos/instrument-repos.service.ts
@@ -1,0 +1,420 @@
+import { execFileSync } from 'node:child_process';
+import { createCipheriv, createDecipheriv, createHash, randomBytes } from 'node:crypto';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
+import * as path from 'node:path';
+
+import { ConfigService, InjectModel, LoggingService } from '@douglasneuroinformatics/libnest';
+import type { Model } from '@douglasneuroinformatics/libnest';
+import { BadGatewayException, ConflictException, Injectable, NotFoundException } from '@nestjs/common';
+import type { OnModuleInit } from '@nestjs/common';
+import { bundle, BUNDLER_FILE_EXT_REGEX, inferLoader } from '@opendatacapture/instrument-bundler';
+import type { BundlerInput } from '@opendatacapture/instrument-bundler';
+
+import { accessibleQuery } from '@/auth/ability.utils';
+import type { EntityOperationOptions } from '@/core/types';
+import { InstrumentsService } from '@/instruments/instruments.service';
+
+import type { CreateInstrumentRepoDto } from './dto/create-instrument-repo.dto';
+
+@Injectable()
+export class InstrumentReposService implements OnModuleInit {
+  constructor(
+    @InjectModel('Group') private readonly groupModel: Model<'Group'>,
+    @InjectModel('InstrumentRepo') private readonly instrumentRepoModel: Model<'InstrumentRepo'>,
+    @InjectModel('Instrument') private readonly instrumentModel: Model<'Instrument'>,
+    private readonly configService: ConfigService,
+    private readonly instrumentsService: InstrumentsService,
+    private readonly loggingService: LoggingService
+  ) {}
+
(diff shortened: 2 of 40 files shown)


thomasbeaudry:
Thanks for the review! Addressed all the CodeRabbit findings (and the earlier CodeQL alert) in 78cea143c.

**Backend**
- **Temp-dir cleanup**: import now removes the exact `tmpDir` it created. Previously, when a repo extracted without a single subdirectory, cleanup ran `rmSync(path.dirname(repoDir))` — i.e. the OS temp root. `downloadAndExtractRepo` now returns `{ repoDir, tmpDir }` and the caller deletes only `tmpDir`.
- **Event-loop blocking**: `curl`/`tar` switched from `execFileSync` to promisified async `execFile`, so imports no longer block the event loop.
- **`buildInstrumentSourceMap`**: scoped to the requested instrument ids and `select`s only the provenance fields, so it no longer loads every record (incl. `bundle`) on each `findInfo`.
- **GROUP_MANAGER access**: `read InstrumentRepo` is now constrained to `{ groupIds: { hasSome: groupIds } }`, matching the Group permission pattern.
- **`groups.updateById` rename check**: now compares the *requested* name (not the current one, which always self-matched) and skips the check when unchanged — fixes the spurious `ConflictException` on rename.

**Frontend**
- **Admin groups**: the edit sheet now closes only after a successful save (selection is preserved on failure), and the mutation also invalidates the instrument-repos cache so the "in use" state refreshes.
- **Localization**: the delete-guard warning is now wrapped in `t({ en, fr })`.
- **List semantics**: the instrument-list dialog's `<ul>` now contains only `<li>` children (header/`<hr>` moved out).
- **A11y**: added an `aria-label` to the icon-only instrument preview button.
- **Stale selection**: added a `useEffect` to resync the group's instrument selection when the upstream group data changes.

**Tests**: added `groups.service` assertions (non-repo-only connect, repo relation set, rename collision) plus `normalizeUrl` cases.

**CodeQL (ReDoS)**: `normalizeUrl` no longer uses the `/\/+$/` regex — trailing slashes are trimmed with a linear scan and `.git` via `endsWith`.

On the one item I did **not** change — moving the import to a worker thread/queue: the async `execFile` change unblocks the event loop, and importing a repo is an infrequent admin-only action, so a worker-thread refactor felt out of scope here. Happy to do it as a follow-up if preferred.

apps/api/src/auth/ability.factory.ts
```diff
@@ -27,6 +27,7 @@ export class AbilityFactory {
         ability.can('manage', 'Assignment', { groupId: { in: groupIds } });
         ability.can('manage', 'Group', { id: { in: groupIds } });
         ability.can('read', 'Instrument');
+        ability.can('read', 'InstrumentRepo', { groupIds: { hasSome: groupIds } });
```
joshunrau, line 30: Doesn't standard user also need to be able to read this?

joshunrau: Otherwise, could cause problems. Did you verify everything works for standard users?

apps/api/src/instrument-repos/instrument-repos.service.ts
```diff
@@ -0,0 +1,437 @@
...
+          '-H',
+          'Accept: application/vnd.github+json',
+          '-H',
+          'X-GitHub-Api-Version: 2022-11-28'
+        ];
+        if (token) {
+          curlArgs.push('-H', `Authorization: Bearer ${token}`);
+        }
+        curlArgs.push(tarballUrl, '-o', tarballPath);
+        await execFileAsync('curl', curlArgs, { timeout: 60_000 });
```
joshunrau: do we know that curl is installed in the docker containers? I don't think it will be. Is there a Node-based way to do this?

apps/web/src/hooks/useCreateInstrumentRepoMutation.ts
```diff
@@ -0,0 +1,27 @@
...
+import { INSTRUMENT_REPOS_QUERY_KEY } from './useInstrumentReposQuery';
+
+export function useCreateInstrumentRepoMutation() {
+  const queryClient = useQueryClient();
+  const addNotification = useNotificationsStore((store) => store.addNotification);
+  return useMutation({
+    mutationFn: ({ data }: { data: CreateInstrumentRepoData }) =>
+      axios.post('/v1/instrument-repos', data, { meta: { disableDefaultErrorNotification: true } }),
+    onError(err) {
+      addNotification({ message: getApiErrorMessage(err, 'Failed to import repository'), type: 'error' });
```
joshunrau: messages should be translated - use the useTranslation hook

apps/web/src/hooks/useCreateInstrumentRepoMutation.ts
```diff
@@ -0,0 +1,27 @@
...
+    },
+    onSuccess() {
+      addNotification({ message: 'Repository imported successfully', type: 'success' });
```
joshunrau: see above

apps/web/src/hooks/useDeleteInstrumentRepoMutation.ts
```diff
@@ -0,0 +1,26 @@
+import { useNotificationsStore } from '@douglasneuroinformatics/libui/hooks';
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+import axios from 'axios';
+
+import { getApiErrorMessage } from '@/utils/error';
+
+import { INSTRUMENT_REPOS_QUERY_KEY } from './useInstrumentReposQuery';
+
+export function useDeleteInstrumentRepoMutation() {
```
joshunrau, line 9: translate messages

apps/web/src/hooks/useSyncInstrumentRepoMutation.ts
```diff
@@ -0,0 +1,29 @@
+import { useNotificationsStore } from '@douglasneuroinformatics/libui/hooks';
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+import axios from 'axios';
+
+import { getApiErrorMessage } from '@/utils/error';
+
+import { INSTRUMENT_REPOS_QUERY_KEY } from './useInstrumentReposQuery';
+
+export function useSyncInstrumentRepoMutation() {
```
joshunrau, line 9: french

apps/web/src/routes/_app/admin/groups/index.tsx
```diff
@@ -1,19 +1,83 @@
...
+    mutationFn: async ({ groupId, instrumentRepoIds }: { groupId: string; instrumentRepoIds: string[] }) => {
+      const response = await axios.patch(
+        `/v1/groups/${groupId}`,
+        { instrumentRepoIds },
+        { meta: { disableDefaultErrorNotification: true } }
+      );
+      return $Group.parseAsync(response.data);
+    },
+    onError(err) {
+      addNotification({ message: getApiErrorMessage(err, 'Failed to update group repositories'), type: 'error' });
```
joshunrau: french

apps/web/src/routes/_app/admin/groups/index.tsx
```diff
@@ -29,6 +93,15 @@ const RouteComponent = () => {
         columns={[
           {
             accessorKey: 'name',
+            cell: (ctx) => {
+              const group = ctx.row.original;
+              return (
+                <span className="flex items-center">
+                  {group.name}
+                  <span className="hidden" data-row-selected={highlightedRowId === group.id ? 'true' : 'false'} />
```
joshunrau, line 107: french

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
```diff
@@ -0,0 +1,384 @@
...
+          en: 'Please enter a valid GitHub repository URL.',
+          fr: 'Veuillez entrer une URL de dépôt GitHub valide.'
+        })
+      );
+      return;
+    }
+    onSubmit(result.data);
+  };
+
+  return (
```
joshunrau, line 64: is there a reason the libui form cannot be used here?

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
```diff
@@ -0,0 +1,384 @@
...
+      <DataTable
+        columns={[
+          {
+            accessorKey: 'name',
+            cell: (ctx) => {
+              const repo = ctx.row.original;
+              return (
+                <span className="flex items-center">
+                  {repo.name}
+                  <span className="hidden" data-row-selected={highlightedRowId === repo.id ? 'true' : 'false'} />
```
joshunrau, line 170: french

apps/web/src/routes/_app/admin/users/index.tsx
```diff
@@ -380,6 +385,15 @@ const RouteComponent = () => {
         columns={[
           {
             accessorKey: 'username',
+            cell: (ctx) => {
+              const user = ctx.row.original;
+              return (
+                <span className="flex items-center">
+                  {user.username}
+                  <span className="hidden" data-row-selected={highlightedRowId === user.id ? 'true' : 'false'} />
```
joshunrau, line 393: french

apps/web/src/routes/_app/datahub/index.tsx
```diff
@@ -363,8 +364,14 @@ const MasterDataTable: React.FC<{
...
             accessorFn: (subject) => removeSubjectIdScope(subject.id),
             cell: (ctx) => {
-              const value = ctx.getValue() as string;
-              return value.slice(0, subjectIdDisplaySetting ?? 9);
+              const subject = ctx.row.original;
+              const value = (ctx.getValue() as string).slice(0, subjectIdDisplaySetting ?? 9);
+              return (
+                <span className="flex items-center">
+                  {value}
+                  <span className="hidden" data-row-selected={highlightedRowId === subject.id ? 'true' : 'false'} />
```
joshunrau, line 372: french

joshunrau (changes requested):
looks great! A few minor comments. Also, French translations missing in quite a few places. Can you please just have the agent review and make sure everything is translted.

apps/api/src/auth/ability.factory.ts
thomasbeaudry: Verified — standard users don't need `read InstrumentRepo`. The provenance shown in the UI (`sourceRepo`) is denormalized onto each Instrument record and served via `/instruments/info`, which is gated only by `read Instrument` (standard users have that unconstrained). `buildInstrumentSourceMap` reads `sourceRepoId`/`sourceRepoName` straight off the Instrument — it never queries `InstrumentRepo` or checks that ability. The `/instrument-repos` endpoints are consumed solely by the two admin-only pages (`useInstrumentReposQuery` is imported only there), so standard users never hit them. I checked the manage-group / datahub / accessible-instruments flows and they work for standard users. Granting standard users repo read would just be over-permissive, so I left the ability as-is.

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
thomasbeaudry: Intentional — this is a plain `<form>` rather than the libui `Form` because of `autoComplete="off"` plus the non-semantic field names, which are needed to stop browsers/password managers from autofilling the URL and access-token fields (there's a comment to that effect above the form). Happy to switch to libui `Form` if you'd prefer and we find another way to suppress autofill.

apps/api/src/instrument-repos/instrument-repos.service.ts
gdevenyi: There's definitely a node-based way to do this.

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
gdevenyi: >  find another way to suppress autofill.

We can add that to libUI if its not already available.

PR #1384, opened 2026-06-23
feat(web): add dedicated Remote Assignment page

thomasbeaudry:
## Summary
- Adds a new **Remote Assignment** page under Session that provides an alternative way to create remote assignments by selecting an instrument from the showcase and specifying an expiry date. This does **not** replace the existing assignment creation flow — it is an additional entry point for the same functionality.
- Updates the assignment result slider to include guidance on recovering a lost link via View Current Subject > Assignments.
- Replaces the `ClientTable` in the assignments datahub view with a custom `Table` implementation that highlights the selected row when clicked.
- Adds e2e test scaffolding for the remote assignment page.

## Test plan
- [ ] Navigate to Session > Remote Assignment and verify the instrument showcase is displayed
- [ ] Select an instrument, set an expiry date, and confirm the assignment is created successfully
- [ ] Verify the result slider shows the link, copy button, QR code, and the updated guidance text about finding a lost link
- [ ] Navigate to View Current Subject > Assignments and verify clicking a row highlights it and opens the assignment slider
- [ ] Verify that the existing assignment creation flow (from the datahub) continues to work as before

## Release Notes

* **New Features**
  * Added a dedicated Remote Assignment page accessible from the session navigation menu when gateway is enabled
  * Remote Assignment form now generates assignment URLs with copy and QR code functionality
  * New instrument selection interface for creating remote assignments

* **Refactor**
  * Assignment creation functionality moved from assignments page to Remote Assignment page
  * Assignments page now displays and manages existing assignments only

* **Tests**
  * Added end-to-end test coverage for Remote Assignment workflow


joshunrau (approved):

apps/web/src/routes/_app/session/remote-assignment.tsx
```diff
@@ -0,0 +1,196 @@
...
+                })
+              }) satisfies z.ZodType<Pick<CreateAssignmentData, 'expiresAt'>>
+            }
+            onSubmit={async ({ expiresAt }) => {
+              const assignment = await createAssignmentMutation.mutateAsync({
+                data: {
+                  expiresAt,
+                  groupId: currentGroup?.id,
+                  instrumentId: selectedInstrument!.id,
+                  subjectId: currentSession.subject!.id
```
gdevenyi, line 174: `currentSession.subject!.id` uses a non-null assertion on `subject`, which the `Session` schema declares as `$Subject.nullable()`. The sibling field `subjectId: z.string()` is guaranteed non-null, so you can drop the assertion entirely:

```ts
subjectId: currentSession.subjectId
```

This is also consistent with `useNavItems.ts`, which reads `currentSession?.subjectId`.

gdevenyi (commented):
## Review — Remote Assignment page

I reviewed the ~400 lines of application code (the rest of the diff is `pnpm-lock.yaml`). The core logic checks out, and I verified the parts that looked risky:

- **`useCreateAssignment` now parses the POST response through `$Assignment`** — sound: the API `create()` returns the full record including `url`, and `$Assignment` includes `url: z.string().url()`, so the parse won't throw.
- **`WithFallback`** correctly suspends until `instrumentInfoQuery.data` is defined, so passing possibly-`undefined` `data` to `InstrumentShowcase` is fine.
- **Selected-row highlight works** — libui's `TableRow` ships `data-[state=selected]:bg-muted` and forwards `data-state`, so `data-state={... ? 'selected' : undefined}` renders the highlight as intended.

The inline notes below are all non-blocking (low / nit): a couple of consistency points, a PR-description/test-plan mismatch worth correcting, and an unused test locator. Nice work overall.

apps/web/src/hooks/useNavItems.ts
```diff
@@ -148,6 +149,15 @@ export function useNavItems() {
         url: '/instruments/accessible-instruments'
       });
     }
+    // Remote assignment requires the gateway to be enabled, since assignments are served through it
+    if (ability?.can('create', 'Assignment') && setupStateQuery.data.isGatewayEnabled) {
```
gdevenyi, line 153: This nav item is now the **only** remaining entry point for creating assignments — the create `Dialog`/`Form` has been removed from `datahub/$subjectId/assignments.tsx`. So when `isGatewayEnabled` is false, there is no longer any UI path to create an assignment.

Functionally the gate is fine (server-side `create()` always calls `gatewayService.createRemoteAssignment`, so creation already required the gateway), but the PR description says this "does **not** replace the existing assignment creation flow" and the test plan asks to verify the datahub flow "continues to work as before" — both are now inaccurate. Please update the description/test plan to reflect that the datahub creation flow was removed and replaced by this page.

apps/web/src/routes/_app/datahub/$subjectId/assignments.tsx
```diff
@@ -96,140 +83,83 @@ const AssignmentsTable: React.FC<{
...
-          label: t('assignments.status')
-        }
-      ]}
-      data={assignments}
-      entriesPerPage={15}
-      minRows={15}
-      onEntryClick={onSelection}
-    />
+    <div className="bg-card text-muted-foreground shadow-xs rounded-md border tracking-tight">
+      <Table>
```
gdevenyi, line 101: Replacing `ClientTable` with a hand-rolled `<Table>` drops the previous `entriesPerPage={15}` / `minRows={15}` behavior, so every assignment now renders at once with no pagination or consistent min-height. `useAssignmentsQuery` fetches all of a subject's assignments unbounded, so a subject with a long history gets an arbitrarily long table. The sibling `upload/index.tsx` still uses `ClientTable` with `entriesPerPage={15}`, so this also diverges from the existing convention. Consider keeping `ClientTable` (it already supports `onEntryClick` and row selection) rather than a custom table.

apps/web/src/routes/_app/session/remote-assignment.tsx
```diff
@@ -0,0 +1,196 @@
...
+  if (!currentSession) {
+    return null;
+  }
+
+  return (
+    <div data-testid="remote-assignment-page">
+      <PageHeader>
+        <Heading className="text-center" variant="h2">
+          {t({
+            en: 'Remote Assignment',
```
gdevenyi, line 106: This page hardcodes all user-facing strings as inline `t({ en, fr })` objects, but the repo convention keeps strings in the translation JSON and references them by key (the sibling `assignments.tsx` in this very PR uses `t('assignments.title')`, `t('assignments.status')`, etc.). Notably this PR *removed* `addAssignment` / `addAssignmentDesc` / `errors.expiryMustBeInFuture` from `datahub.json` and re-inlined equivalent strings here.

Concretely, this H2 (`'Remote Assignment'` / `'Tâche à distance'`) duplicates the nav label you just added at `layout.navLinks.remoteAssignment` in `translations/layout.json`. Recommend moving these strings into the translation JSON and reusing `t('layout.navLinks.remoteAssignment')` here so the label isn't defined twice.

testing/e2e/src/pages/remote-assignment.page.ts
```diff
@@ -0,0 +1,30 @@
...
+  readonly instrumentShowcase: Locator;
+  readonly pageHeader: Locator;
+  readonly resultSlider: Locator;
+
+  constructor(page: Page) {
+    super(page);
+    this.pageHeader = page.getByTestId('page-header');
+    this.instrumentShowcase = page.getByTestId('instrument-showcase');
+    this.createDialog = page.getByRole('dialog');
+    this.resultSlider = page.locator('[data-testid="remote-assignment-page"] ~ div [role="dialog"]');
```
gdevenyi, line 16: `resultSlider` is declared here but never referenced anywhere in the e2e suite — the spec verifies the result panel via `page.getByRole('link', { name: 'Assignment Link' })` and `page.locator('input[readonly]')` instead. It also looks like a broken selector: `AssignmentResultSlider` is a Radix `Sheet`, which portals its content to `document.body`, so `[role="dialog"]` won't be a descendant of a `~ div` sibling of `[data-testid="remote-assignment-page"]`. Suggest removing the unused `resultSlider` locator (and its field on line 9).

apps/web/src/routes/_app/session/remote-assignment.tsx
joshunrau: not relevant both do the same thing

apps/web/src/routes/_app/datahub/$subjectId/assignments.tsx
joshunrau: this is fine, I thought I replaced client table already, must have missed this one or forgot to merge that change.

joshunrau: client table is highly flawed

apps/web/src/hooks/useNavItems.ts
joshunrau: @thomasbeaudry

testing/e2e/src/pages/remote-assignment.page.ts
joshunrau: @thomasbeaudry

Issue #1317, opened 2026-03-19
Change how a user creates a remote assignment


thomasbeaudry:
### Current Implementation and Limitations

The current workflow to start an assignment is non-intuitive.  Having to click on "view current subject"->assignments-"add assignemnt" can be improved on.  Also, when you finally get to select an instrument, you are presented with a simple drop down box that doesn't give you all the of the instrument card info that you get when you administer an instrument (edition,languages,tags,etc)

### Associated Application Components

Client

### Proposed Solution

There are 2 ways to improve this, I lean towards the first idea, but both would be an improvement.

Option #1
After clicking on "Administer Instrument"  you see all of the instrument cards, and when you click on them you are presented with the overview page.  Before this page gets shown, I would have a "Type of session" page with 3 big buttons:  In-person, Remote Assignment, and Retrospective.  This admisitration type should also be recorded in the DB somewhere so we can reference it in the future.  It could also be an optional column in the data hub section if you want.   If you choose this option then you can remove this field on the start session page:

<img width="1263" height="246" alt="Image" src="https://github.com/user-attachments/assets/77fee790-59ca-4b7e-a399-4ebce9a211e5" />

Option #2
Add an option to the menu bar under the "Administer Instrument" called "Remote Assignment". 

<img width="269" height="619" alt="Image" src="https://github.com/user-attachments/assets/ce70d1ce-b3e9-4979-bb33-996ce6eeed8e" />

 This would lead to a near identical page as the Administer Instrument with all of the instrument cards to choose from.

### Estimated Difficulty

Low

### Priority

Medium

PR #1386, opened 2026-06-25
feat(web): add dedicated Remote Assignment page (re-land of #1384)

thomasbeaudry:
Re-lands #1384, which was reverted in #1385, with review feedback addressed.

## What this does
- Adds a dedicated **Remote Assignment** page at `/session/remote-assignment` for creating remote assignments (choose an instrument, set an expiry, get a shareable link / QR code).
- **Removes** the inline "Add assignment" `Dialog`/`Form` from `datahub/$subjectId/assignments.tsx`. Assignment **creation** now lives solely on the dedicated page; the datahub assignments tab now only **displays** existing assignments (view / cancel).
- Adds e2e coverage (`2.4-remote-assignment.spec.ts` + page object).

## Gateway gating (corrected from the original PR description)
The original #1384 description claimed this did *not* replace the existing creation flow and that the datahub flow "continues to work as before" — that was inaccurate and contributed to the revert. To be precise:

When `isGatewayEnabled` is **false**, there is **no assignment UI of any kind** — same behavior as before this feature. Both entry points are gated:
- Remote Assignment nav item — `useNavItems.ts` (behind `isGatewayEnabled`)
- Datahub "Assignments" tab — `datahub/$subjectId/route.tsx` (behind `config.setup.isGatewayEnabled`)

This is intentional and confirmed: server-side `create()` always goes through `gatewayService.createRemoteAssignment`, so assignment creation already required the gateway.

## Changes since #1384
- Removed an unused, broken `resultSlider` locator from the e2e page object (the `AssignmentResultSlider` Radix Sheet portals to `document.body`, so the relative selector never matched and it was never referenced).

## Test plan
- With gateway **enabled**: Remote Assignment nav item and datahub Assignments tab are visible; can create an assignment from the dedicated page and view/cancel it from the datahub tab.
- With gateway **disabled**: neither the nav item nor the datahub Assignments tab appears; no path to create or view assignments.

Refs #1317

## Summary by CodeRabbit

* **New Features**
  * Added a Remote Assignment session flow to create assignments from an active session, including instrument selection, an expiration date, and a results panel with an assignment link (copy + QR code).
  * Added a “Remote Assignment” navigation link when available.
* **Bug Fixes**
  * Improved assignment creation reliability by validating the server response before using it.
  * Updated the Datahub assignments page to focus on viewing/canceling existing assignments.
* **Tests**
  * Added end-to-end coverage for the remote assignment flow, including URL/link assertions.


thomasbeaudry:
@joshunrau this re-lands #1384 with the feedback addressed:

- **Gateway gating** — confirmed `isGatewayEnabled === false` hides all assignment UI (nav item + datahub Assignments tab), matching the "same as before" behavior we discussed. No code change was needed; the existing gates already do this.
- **Dead e2e locator** — removed the unused/broken `resultSlider` locator you flagged.
- **PR description** — rewritten to accurately state the datahub create flow was removed and creation now lives on the dedicated page (the original description was the misleading part).

Ready for another look when you have a sec.

PR #1392, opened 2026-07-06
Add series instrument creation and remote assignment flow

thomasbeaudry:
Includes the create-series-instrument feature (bundler, schemas, renderer, web hooks/routes, e2e coverage) and a fix for the group-save 500: filter deleted instrument ids server-side before the relation set, and prune them from the cached group after an instrument delete.

Something I noticed is that the datahub doesn't track whether an insturment is administered standalone or within a series.  It may be useful to track this because I could imagine the results of a form could vary depending on the other forms they take at the same time?  Also the datahub records the form date, but no the form adminsitration time.  

Do you want me to add those 2 things to the data hub?  If we determine that is it important ot track the series in the DB, then i would disallow deletiion of a series if it has been adminsitered to a participant for reproducibility reasons.


gdevenyi:
> Do you want me to add those 2 things to the data hub? If we determine that is it important ot track the series in the DB, then i would disallow deletiion of a series if it has been adminsitered to a participant for reproducibility reasons.

We need to review the "metadata" model of instruments in general to make sure we're capturing all we should capture.

joshunrau:
> > Do you want me to add those 2 things to the data hub? If we determine that is it important ot track the series in the DB, then i would disallow deletiion of a series if it has been adminsitered to a participant for reproducibility reasons.
> 
> We need to review the "metadata" model of instruments in general to make sure we're capturing all we should capture.

The data hub does not track anything. The data hub is just displaying the results from the database. We record both the exact time that an instrument was completed as well as the mode of administration (remote, in-person, retrospective). Both of these should be available when you download the results as a CSV.

I agree that we should display the mode of capture. I did not add exact time to the table to avoid cluttering it. If you would prefer, we could add something to display the time, I just didn't see the use case.

gdevenyi:
> If you would prefer, we could add something to display the time, I just didn't see the use case.

I'd think I'd like to discuss how we can make customiziable columns in the viewer so that data can be available if wanted. Seperate issue to this (and to the broader metadata audit)

thomasbeaudry:
> I'd think I'd like to discuss how we can make customiziable columns in the viewer so that data can be available if wanted. Seperate issue to this (and to the broader metadata audit)

Yes i was thinking the exact same thing.  

  @joshunrau also I downloaded the happiness questionaire csv and I don't see the  mode of administration:
image

Also, is the series that was used to record the data being recorded in the DB?  This would be linked to this PR if we want ot track it.

apps/api/src/auth/ability.factory.ts
```diff
@@ -27,6 +27,10 @@ export class AbilityFactory {
         ability.can('manage', 'Assignment', { groupId: { in: groupIds } });
         ability.can('manage', 'Group', { id: { in: groupIds } });
         ability.can('read', 'Instrument');
+        // Group managers may assemble series instruments on the fly and remove ones they created
+        // (the service enforces that an instrument with collected records cannot be deleted).
+        ability.can('create', 'Instrument');
+        ability.can('delete', 'Instrument');
```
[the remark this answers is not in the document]
thomasbeaudry: Addressed on the current branch: the delete ability is now scoped — `ability.can('delete', 'Instrument', { sourceRepoId: null })` (ability.factory.ts:33). `accessibleQuery` therefore no longer returns `{}`, so a group manager can only delete non-repo (manually-uploaded) instruments; repo-sourced instruments shared across groups are excluded. The service-side record-count guard remains as defense-in-depth.

apps/api/src/instruments/instruments.service.ts
```diff
@@ -106,6 +120,74 @@ export class InstrumentsService {
...
+   * with it (so historical data is never orphaned). Also detaches it from any group that exposes it.
+   */
+  async deleteById(id: string, { ability }: EntityOperationOptions = {}): Promise<{ id: string }> {
+    const instrument = await this.instrumentModel.findFirst({
+      where: { AND: [accessibleQuery(ability, 'delete', 'Instrument')], id }
+    });
+    if (!instrument) {
+      throw new NotFoundException(`Failed to find instrument with ID: ${id}`);
+    }
+    const recordCount = await this.instrumentRecordModel.count({ where: { instrumentId: id } });
```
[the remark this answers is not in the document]
thomasbeaudry: `deleteById` gates on `accessibleQuery(ability, 'delete', 'Instrument')`, and the delete ability now carries `{ sourceRepoId: null }` (ability.factory.ts:33), so the generated Prisma `where` already excludes repo-sourced instruments — a repo instrument yields no match and returns NotFound. Happy to also add an explicit `sourceRepoId === null` assertion in the service as belt-and-suspenders (independent of the ability config) if you'd like it guaranteed at the service layer too.

apps/api/src/instruments/instruments.service.ts
```diff
@@ -178,10 +260,18 @@ export class InstrumentsService {
...
 
     const sourceMap = await this.buildInstrumentSourceMap(instances.map((instance) => instance.id));
+    const scalarInstrumentIds = new Map(
+      instances.flatMap((instance) => {
+        if (!isScalarInstrument(instance) || !instance.internal) {
+          return [];
+        }
+        return [[`${instance.internal.name}:${instance.internal.edition}`, instance.id] as const];
+      })
+    );
```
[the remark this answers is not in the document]
thomasbeaudry: Already handled: `scalarInstrumentIds` (and `sourceMap`) are built from `allInstances` — the unfiltered fetch — not the kind-filtered `instances` (instruments.service.ts:265-272). So when `findInfo` is called with `kind=SERIES`, the scalar id lookup is still complete and series can resolve their scalar items.

apps/web/src/routes/_app/group/manage.tsx
```diff
@@ -490,7 +878,9 @@ const RouteComponent = () => {
         authors: instrument.details.authors,
         description: instrument.details.description,
         id: instrument.id,
+        internal: instrument.internal ?? null,
         kind: instrument.kind,
+        seriesItems: instrument.seriesItems,
         source,
```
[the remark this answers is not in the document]
thomasbeaudry: Addressed: `source` is now derived from `repoId`, not the repo name — `const source = repoId ? { kind: 'repo', name: repoName ?? 'Unknown' } : { kind: 'manual' }` (manage.tsx:876). A legacy repo instrument with a null name is therefore still treated as repo-sourced (so it doesn't get the manual/delete affordance), with an "Unknown" fallback label.

apps/api/src/auth/ability.factory.ts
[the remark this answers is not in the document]
joshunrau: Why do group managers need to be able to delete instruments in the first place? This contradicts the original design of the platform.

apps/api/src/instruments/__tests__/instruments.service.spec.ts
```diff
@@ -2,32 +2,264 @@ import { CryptoService, getModelToken, LoggingService, VirtualizationService } f
...
 
 import { InstrumentsService } from '../instruments.service';
 
+// Avoid bundling a real instrument (esbuild) during unit tests; `createSeries` only needs `create` to be
+// called with whatever the bundler produces.
+vi.mock('@opendatacapture/instrument-bundler', () => ({
+  bundle: vi.fn(() => Promise.resolve('__BUNDLE__'))
+}));
+
+// A multilingual series instance as returned by `find`, containing two forms.
```
joshunrau, line 20: add proper type annotation

apps/api/src/instruments/dto/create-series-instrument.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { ValidationSchema } from '@douglasneuroinformatics/libnest';
+import type { ScalarInstrumentInternal } from '@opendatacapture/runtime-core';
+import { $CreateSeriesInstrumentData } from '@opendatacapture/schemas/instrument';
+
+@ValidationSchema($CreateSeriesInstrumentData)
```
joshunrau: no need for dto anymore. just import the $CreateSeriesInstrumentData any use that as annotation. make sure not to use type-only import

apps/api/src/instruments/dto/create-series-instrument.dto.ts
```diff
@@ -0,0 +1,12 @@
...
+export class CreateSeriesInstrumentDto {
+  confirmDuplicate?: boolean;
+  description?: string;
+  instructions?: string;
+  items: ScalarInstrumentInternal[];
+  title: string;
+}
```
joshunrau: Why are you not using the actual series instrument definition? Right now, you are missing the multilingual aspects and potential other things. Use the same data structure as the instrument type, omitting things that are not stored/computed

apps/api/src/instruments/instruments.service.ts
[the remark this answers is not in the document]
joshunrau: the ability should be scoped to only series instruments and this should be explicitly tested.

apps/api/src/instruments/instruments.service.ts
```diff
@@ -106,6 +120,74 @@ export class InstrumentsService {
...
+   */
+  async deleteById(id: string, { ability }: EntityOperationOptions = {}): Promise<{ id: string }> {
+    const instrument = await this.instrumentModel.findFirst({
+      where: { AND: [accessibleQuery(ability, 'delete', 'Instrument')], id }
+    });
+    if (!instrument) {
+      throw new NotFoundException(`Failed to find instrument with ID: ${id}`);
+    }
+    const recordCount = await this.instrumentRecordModel.count({ where: { instrumentId: id } });
+    if (recordCount > 0) {
```
joshunrau, line 219: records are never created for series instruments, only for constituent members

apps/api/src/instruments/instruments.service.ts
```diff
@@ -281,6 +378,75 @@ export class InstrumentsService {
     return map;
   }
 
+  /**
+   * The title of an existing series instrument that contains the exact same set of forms as `items`
+   * (regardless of order), or `null` when none exists. Used to warn before creating a redundant series.
+   */
+  private async findSeriesTitleWithSameForms(
```
joshunrau: rethink this as series instrument title is not always a string

apps/api/src/instruments/instruments.service.ts
```diff
@@ -281,6 +378,75 @@ export class InstrumentsService {
...
+    const definition = {
+      __runtimeVersion: 1,
+      ...(instructions && instructions.length > 0
+        ? { clientDetails: { instructions: { en: [instructions], fr: [instructions] } } }
+        : {}),
+      content: {
+        items: items.map(({ edition, name }) => ({ edition, name }))
+      },
+      details: {
+        description: { en: text, fr: text },
```
joshunrau: ...

joshunrau: will be adding Claude.md to account for these repeated mistakes

apps/api/src/instruments/instruments.service.ts
```diff
@@ -281,6 +378,75 @@ export class InstrumentsService {
...
+    description?: string;
+    instructions?: string;
+    items: ScalarInstrumentInternal[];
+    title: string;
+  }): string {
+    const text = description && description.length > 0 ? description : title;
+    const definition = {
+      __runtimeVersion: 1,
+      ...(instructions && instructions.length > 0
+        ? { clientDetails: { instructions: { en: [instructions], fr: [instructions] } } }
```
joshunrau: translate all

apps/api/src/instruments/instruments.service.ts
```diff
@@ -281,6 +378,75 @@ export class InstrumentsService {
...
+      content: {
+        items: items.map(({ edition, name }) => ({ edition, name }))
+      },
+      details: {
+        description: { en: text, fr: text },
+        license: 'UNLICENSED',
+        title: { en: title, fr: title }
+      },
+      kind: 'SERIES',
+      language: ['en', 'fr'],
```
joshunrau: this needs to be configurable

apps/api/src/instruments/instruments.service.ts
```diff
@@ -281,6 +378,75 @@ export class InstrumentsService {
...
+    items,
+    title
+  }: {
+    description?: string;
+    instructions?: string;
+    items: ScalarInstrumentInternal[];
+    title: string;
+  }): string {
+    const text = description && description.length > 0 ? description : title;
+    const definition = {
```
joshunrau: there needs to be an explicit type annotation here so this doesn't get out of sync

apps/api/src/instruments/instruments.service.ts
```diff
@@ -281,6 +378,75 @@ export class InstrumentsService {
...
+      tags: { en: ['Series'], fr: ['Série'] }
+    };
+    return `export default ${JSON.stringify(definition)};`;
+  }
+
+  /**
+   * A stable, order-independent key identifying the set of forms in a series. Two series with the same
+   * forms produce the same key even if the forms were added in a different order.
+   */
+  private getFormSetKey(items: ScalarInstrumentInternal[]): string {
```
joshunrau: this is a terrible name - series instruments do not contain only forms

apps/api/src/instruments/instruments.service.ts
```diff
@@ -289,6 +455,20 @@ export class InstrumentsService {
     );
   }
 
+  /**
+   * Whether an instrument visible to the caller already uses the given title (case-insensitive). Used to
+   * enforce that newly-created series instruments have a unique, recognizable name.
+   */
+  private async isInstrumentTitleTaken(title: string, { ability }: EntityOperationOptions = {}): Promise<boolean> {
+    const normalized = title.trim().toLowerCase();
+    const instances = await this.find({}, { ability });
```
joshunrau: now, you are not searching all of the instruments, only the ones the creating user has access to. This seems like a poor idea

apps/api/src/instruments/instruments.service.ts
```diff
@@ -289,6 +455,20 @@ export class InstrumentsService {
...
+    return instances.some((instance) => {
+      const value = instance.details.title;
+      const titles = typeof value === 'string' ? [value] : Object.values(value);
```
joshunrau: so we are just ignoring all non-string titles? lol

apps/web/src/hooks/useCreateSeriesInstrumentMutation.ts
```diff
@@ -0,0 +1,59 @@
...
+import axios from 'axios';
+
+import { getApiErrorMessage } from '@/utils/error';
+
+type CreateSeriesInstrumentInput = {
+  confirmDuplicate?: boolean;
+  description?: string;
+  instructions?: string;
+  items: ScalarInstrumentInternal[];
+  title: string;
```
joshunrau: totally rewrite in accordance with other comment. This should not be duplicated. Use the type that already exists, or create a single new one picking/omitting from that.

apps/web/src/hooks/useCreateSeriesInstrumentMutation.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+// The server either creates the instrument (returning it with an id) or, when another series already
+// contains the same forms and creation was not confirmed, asks the client to confirm the duplicate.
+type SeriesDuplicateConfirmation = {
```
joshunrau: DRY

apps/web/src/hooks/useCreateSeriesInstrumentMutation.ts
```diff
@@ -0,0 +1,59 @@
...
+  existingTitle: string;
+  requiresConfirmation: true;
+};
+type CreateSeriesInstrumentResponse = SeriesDuplicateConfirmation | { id: string };
+
+export function isSeriesDuplicateConfirmation(
+  response: CreateSeriesInstrumentResponse
+): response is SeriesDuplicateConfirmation {
+  return 'requiresConfirmation' in response && response.requiresConfirmation;
```
joshunrau: horrible approach. create a shared type and enforce the contract. use discriminated unions not this flaky approach

apps/web/src/hooks/useDeleteInstrumentMutation.ts
```diff
@@ -0,0 +1,32 @@
...
+          err,
+          t({ en: 'Failed to delete instrument', fr: "Échec de la suppression de l'instrument" })
+        ),
+        type: 'error'
+      });
+    },
+    onSuccess(_, { id }) {
+      addNotification({ type: 'success' });
+      // Drop the deleted id from the cached group so the manage page does not re-send it on save.
+      pruneDeletedInstrument(id);
```
joshunrau: consider whether this is the optimal design

apps/web/src/store/types.ts
```diff
@@ -16,6 +16,7 @@ export type AuthSlice = {
   currentUser: CurrentUser | null;
   login: (accessToken: string) => void;
   logout: () => void;
+  pruneDeletedInstrument: (instrumentId: string) => void;
```
joshunrau: this is horrible. Why the hell is this being placed in the auth store?

apps/web/src/store/slices/auth.slice.ts
```diff
@@ -39,6 +39,19 @@ export const createAuthSlice: SliceCreator<AuthSlice> = (set) => {
     },
     logout: () => {
       window.location.reload();
+    },
+    pruneDeletedInstrument: (instrumentId) => {
```
joshunrau: this is a terrible design. I understand why it might be necessary, but consider how you can change implementation details to make a clean approach

packages/react-core/src/components/InstrumentRenderer/ScalarInstrumentRenderer.tsx
```diff
@@ -37,6 +37,7 @@ export type ScalarInstrumentRendererProps = {
   /** @deprecated */
   options?: InterpretOptions;
   subject?: SubjectDisplayInfo;
+  submitButtonLabel?: string;
```
joshunrau: french

packages/react-core/src/components/InstrumentRenderer/InstrumentRenderer.tsx
```diff
@@ -21,6 +21,7 @@ export type InstrumentRendererProps = {
   NavigationBlocker?: NavigationBlockerComponent;
   onSubmit: InstrumentSubmitHandler;
   subject?: SubjectDisplayInfo;
+  submitButtonLabel?: string;
```
joshunrau: french

packages/schemas/src/instrument/instrument.base.ts
```diff
@@ -272,6 +280,20 @@ const $CreateInstrumentData = z.object({
...
+  // When true, proceed even though another series already contains the same set of forms.
+  confirmDuplicate: z.boolean().optional(),
+  // Optional human-readable description; falls back to the title when omitted.
+  description: z.string().min(1).optional(),
+  // Optional instructions shown to participants before the series begins.
+  instructions: z.string().min(1).optional(),
+  // The ordered list of scalar instruments (by name + edition) that make up the series.
+  items: z.array($ScalarInstrumentInternal).min(2),
+  // The unique display name entered by the group manager.
+  title: z.string().min(1)
```
joshunrau: this should be your source of truth and should come based on the master schema

packages/schemas/src/instrument/instrument.base.ts
```diff
@@ -248,6 +250,12 @@ const $InstrumentInfo = $BaseInstrument
   .extend({
     id: z.string(),
     internal: $ScalarInstrumentInternal.optional(),
+    seriesItems: z
+      .object({
+        id: z.string()
+      })
+      .array()
+      .optional(),
     sourceRepo: z
```
joshunrau: we should be using a discriminated union here for series vs scalar

joshunrau (changes requested):
this is terrible, it needs major changes

apps/api/src/instruments/__tests__/instruments.service.spec.ts
thomasbeaudry: @joshunrau   a series instrument isn't really an instrument, it's a bundled list of instruments.  I see no harm in letting the people who create them delete them,  They could even belong to speccific groups instead of every user as well

apps/api/src/auth/ability.factory.ts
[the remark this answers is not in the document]
thomasbeaudry: Addressed in the latest push. The rationale for keeping it: a series instrument isn't a shared platform asset — it's an on-the-fly bundle of existing instruments that a group manager assembles, and records are only ever collected against the constituent scalars, never the series itself. So deleting one never orphans data or touches a real instrument. The ability is now scoped — `ability.can('delete', 'Instrument', { sourceRepoId: null })` — so repo-sourced instruments can never be deleted, and the service additionally rejects deletion of anything that isn't a series (unit-tested). If you'd still rather group managers not have this at all, I can drop the delete path entirely.

apps/api/src/instruments/__tests__/instruments.service.spec.ts
thomasbeaudry: Done — `existingSeries` is now typed as `WithID<SeriesInstrument>` (no `as any`), which also forced the fixture to carry the full details/license/tags shape.

apps/api/src/instruments/dto/create-series-instrument.dto.ts
thomasbeaudry: Done — the DTO class is deleted. `$CreateSeriesInstrumentData` is exported as both the schema value and its inferred type, and the controller uses it directly as the `@Body()` annotation (value import, not type-only) — same pattern as `$SelfUpdateUserData` in the users controller.

apps/api/src/instruments/dto/create-series-instrument.dto.ts
thomasbeaudry: Reworked — the request schema is now composed from the master instrument schemas: `details` picked from `$InstrumentDetails` (multilingual title + optional description), `clientDetails.instructions` from `$ClientInstrumentDetails`, plus `language: $InstrumentLanguage`. Only fields that are stored verbatim are accepted; everything computed server-side (id, content, kind) is omitted. The generated definition is explicitly typed `Omit<SeriesInstrument, 'id'>` so any drift from the instrument type fails to compile.

apps/api/src/instruments/instruments.service.ts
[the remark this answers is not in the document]
thomasbeaudry: Done in the latest push — `deleteById` now rejects anything that isn't a series (`isSeriesInstrument` check → ForbiddenException), with an explicit unit test ('refuses to delete a non-series (scalar) instrument'). One caveat on scoping the *ability* to series: `kind` isn't a database column (it lives inside the evaluated bundle), so the CASL/Prisma layer can only scope on real fields — the ability is scoped to `{ sourceRepoId: null }` and the series-only rule is enforced (and tested) at the service layer.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Right — the record-count guard is removed entirely. `deleteById` no longer touches `instrumentRecordModel`; it only permits series (which never have records of their own), and the test asserts `count` is never called.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Fixed — titles are handled in their stored `InstrumentUIOption` form everywhere: `isInstrumentTitleTaken` compares every language value (case-insensitively), and the duplicate check returns the stored multilingual title as-is so the client localizes it.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: This block is rewritten — the definition is now explicitly typed `Omit<SeriesInstrument, 'id'>` and built verbatim from the caller's multilingual details and language. No hardcoded en/fr duplication, no untyped JSON blob.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Done — nothing is duplicated across languages anymore. Details/instructions are stored exactly as authored in the request's `language`, and the generated tags are localized per language ('Series'/'Série').

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Done — `language` is now part of the request (`$InstrumentLanguage`, unilingual or multilingual) and the details take the corresponding shape. Nothing about the series is hardcoded server-side anymore.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Done — `const definition: Omit<SeriesInstrument, 'id'> = ...`, so a change to the instrument shape fails to compile here instead of drifting silently.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Renamed — `getFormSetKey` → `getSeriesItemSetKey`, and `findSeriesTitleWithSameForms` → `findSeriesTitleWithSameItems`. Comments and messages no longer say 'forms' when they mean series items.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Fixed — `isInstrumentTitleTaken` no longer applies the caller's ability filter; uniqueness is checked against the full instrument set, since a title must be globally unique rather than unique among what the creator happens to see.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Fixed — every language value of every title is compared (case-insensitive), via a small `titleValues` helper that flattens `string | multilingual object` uniformly. Nothing is skipped anymore.

apps/web/src/hooks/useCreateSeriesInstrumentMutation.ts
thomasbeaudry: Rewritten — the hook now uses the shared contract from `@opendatacapture/schemas/instrument`: `$CreateSeriesInstrumentData` as the request body and `CreateSeriesInstrumentResult` as the response. No locally duplicated types, so client and server can't drift.

apps/web/src/hooks/useCreateSeriesInstrumentMutation.ts
thomasbeaudry: Gone — both request and response types come from the shared schemas package now.

apps/web/src/hooks/useCreateSeriesInstrumentMutation.ts
thomasbeaudry: Done — the response is `CreateSeriesInstrumentResult`, a discriminated union over `outcome` (`'created' | 'duplicate'`) shared between the service, controller, and this hook. The structural `isSeriesDuplicateConfirmation` check is deleted.

apps/web/src/hooks/useDeleteInstrumentMutation.ts
thomasbeaudry: Reworked — the mutation no longer reaches into the auth store. The server already detaches the deleted instrument from every group; the manage page tracks just-deleted ids as local component state and filters them from the save payload, so the stale-id problem is solved where it actually exists.

apps/web/src/store/types.ts
thomasbeaudry: Removed — `pruneDeletedInstrument` is gone from the store entirely; the auth slice is back to auth concerns only.

apps/web/src/store/slices/auth.slice.ts
thomasbeaudry: Removed. The underlying need (don't re-send a deleted id on the next group save) is handled locally in the manage route: a `deletedIds` set in component state, filtered at submit time. No global store involvement.

packages/react-core/src/components/InstrumentRenderer/ScalarInstrumentRenderer.tsx
thomasbeaudry: Changed — `submitButtonLabel` is now a `LocalizedText` (`{ [L in Language]?: string }`) resolved with `t()` inside `FormContent`, so the shared component can no longer be handed a pre-translated single-language string.

packages/react-core/src/components/InstrumentRenderer/InstrumentRenderer.tsx
thomasbeaudry: Same fix as on ScalarInstrumentRenderer — the prop is `LocalizedText` on all three renderers and translated at the leaf with `t()`.

packages/schemas/src/instrument/instrument.base.ts
thomasbeaudry: Done — `$CreateSeriesInstrumentData` is now composed from the master schemas (`$InstrumentDetails`, `$ClientInstrumentDetails`, `$InstrumentLanguage`, `$ScalarInstrumentInternal`) rather than redeclaring its own shapes.

packages/schemas/src/instrument/instrument.base.ts
thomasbeaudry: Done — `InstrumentInfo` is now a discriminated union on `kind`: scalar variants carry a required `internal`, the series variant carries required `seriesItems`. `findInfo` builds the proper variant and consumers narrow on `kind` instead of optional-chaining.

apps/api/src/instruments/instruments.service.ts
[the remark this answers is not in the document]
thomasbeaudry: Following up — my earlier reply was wrong: the lookup was in fact built from the kind-filtered list. Now actually fixed: when a `kind` filter would exclude scalars but the result contains series, `findInfo` builds the scalar-id lookup from the unfiltered instrument set, so `seriesItems` resolves correctly for `kind=SERIES` queries.

apps/web/src/routes/_app/group/manage.tsx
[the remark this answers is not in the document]
thomasbeaudry: Following up — correcting my earlier reply: the committed code was still deriving provenance from the repo name. Now fixed: `source` is keyed on the repo id, with a localized 'Unknown repository' fallback label, so a legacy repo instrument with a null name is still treated as repo-sourced and never gets the manual/delete affordance.

gdevenyi (commented):
## Review: Add series instrument creation and remote assignment flow

Comprehensive pass over `c2478b9`. Overall the rework in the last commit is a big improvement — the schemas are clean (discriminated unions on `kind` and `outcome`, `$CreateSeriesInstrumentData` composed from the master schemas, explicit `Omit<SeriesInstrument,'id'>` typing), the service-layer guards are sound, test coverage for the new service paths is solid, and all user-facing strings are localized. CI (lint-and-test, CodeQL, Analyze) is green. Posting as **Comment** rather than Approve because of the permission expansion and the still-open design sign-off from @joshunrau.

---

### 🔴 Needs decision before merge

**1. `GROUP_MANAGER` now has unscoped `create` on `Instrument` — this also unlocks the regular scalar-upload endpoint.** `apps/api/src/auth/ability.factory.ts:34`
On `main`, `GROUP_MANAGER` only had `read` on `Instrument`. This PR adds `ability.can('create', 'Instrument')` (unscoped). That's required so the series route passes `@RouteAccess({ action: 'create' })`, but `POST /instruments` (the **regular** scalar instrument-upload endpoint, `instruments.controller.ts:26`) is gated by the exact same action/subject. Net effect: group managers can now upload **arbitrary scalar instrument bundles** (code executed in the virtualization sandbox) platform-wide — previously admin-only — and those instruments are visible to everyone. This looks incidental to the series feature. If series-only is intended, consider a dedicated subject/action for series creation (or otherwise keep the regular create admin-only) so the ability doesn't over-grant.

**2. Group-manager `delete` on `Instrument` — still needs @joshunrau's explicit sign-off.** `apps/api/src/auth/ability.factory.ts:35`
@joshunrau's earlier review (changes requested) explicitly questioned why group managers can delete instruments at all ("contradicts the original design"). The latest push scopes it to `{ sourceRepoId: null }` and the service layer enforces series-only with a test (`instruments.service.spec.ts:261`) — good defense-in-depth. But two things to confirm:
- The ability itself still permits delete on **any** manually-uploaded scalar (`sourceRepoId: null`); only the service-level `isSeriesInstrument` check (instruments.service.ts:166) prevents it. If that service check is ever refactored away, manual scalars become deletable by managers. The CASL layer can't narrow to series because `kind` isn't a DB column — acknowledged — so please get explicit maintainer buy-in that this trade-off is acceptable (or drop the delete path and have admins delete series).
- This is a genuine design disagreement that appears resolved-in-code but not yet approved. Want @joshunrau to ack before this merges.

---

### 🟡 Medium

**3. Duplicate-items detection is ability-scoped, while title uniqueness is global — inconsistent.** `instruments.service.ts:399`
`isInstrumentTitleTaken` correctly searches the full instrument set (`find({})`, no ability), but `findSeriesTitleWithSameItems` calls `this.find({ kind: 'SERIES' }, { ability })`. So a manager won't be warned about an identical series built by someone in another group (different title → passes both checks → near-duplicate series created). Either make the duplicate-set check global too (consistent with the title check) or document why it's intentionally scoped.

**4. Creating/deleting a series discards pending instrument toggles in the manage form.** `apps/web/src/routes/_app/group/manage.tsx:602`
Both new flows invalidate `['instrument-info']` (hooks at `useCreateSeriesInstrumentMutation.ts:38` and `useDeleteInstrumentMutation.ts:27`). The refetch recomputes `data` → new `initialSelectedIds` reference → the (pre-existing) `useEffect` at `manage.tsx:602` resets `selectedIds`. If a user has toggled several instruments and then creates or deletes a series, those pending checkbox changes are silently lost. Pre-existing pattern, but newly triggerable. Consider not invalidating until the dialog closes, or guarding the resync while there are unsaved edits.

**5. `DELETE /instruments/:id` is series-only in disguise.** `instruments.controller.ts:43`
The service throws `ForbiddenException` for any non-series instrument — including for admins. The Swagger summary "Delete Instrument" implies a general delete. Either route it as `/instruments/series/:id` or note the series-only contract in the summary/description so callers aren't surprised that it 403s on scalars.

---

### 🟢 Minor / nits

**6. `pnpm-lock.yaml` — 8640 lines of formatting-only churn, zero `package.json` changes.** The diff reduces to "no syntactic changes", i.e. a reformat from a different pnpm patch version with no dependency movement. Regenerate against the repo's pinned pnpm so the PR diff stays reviewable and we avoid future merge conflicts.

**7. Type-naming convention: `$CreateSeriesInstrumentData` is `$`-prefixed as a *type*.** `packages/schemas/src/instrument/instrument.base.ts:307`
The sibling right above it follows `CreateInstrumentData` (type, no `$`) / `$CreateInstrumentData` (value). The new one reuses the `$` name for both type and value. Works, but inconsistent — consider naming the inferred type `CreateSeriesInstrumentData`.

**8. Cost of full-set scans.** `instruments.service.ts:468` / `:394` / `:262`
Each `createSeries` triggers `find({})` (all instruments, full virtualization) for the title check, plus `find({ kind: 'SERIES' })` for the duplicate check; `findInfo` now does an extra full `find({})` whenever the result contains series (`:262`). Fine at current scale — flagging for awareness if the instrument count grows.

**9. Missing regression tests for two specifically-discussed fixes.**
- `findInfo` `seriesItems` resolution under `kind=SERIES` (the Copilot-flagged bug fixed in `:262`) — no test.
- No `ability.factory` test asserting group managers can't delete repo instruments (the scoping rationale). The service-level series-only refusal is covered; the ability scoping itself isn't.

**10. PR description says "e2e coverage"** but the diff only adds the two unit specs under `apps/api/.../__tests__/` — no `testing/` e2e files. Minor description mismatch.

---

### What's done well
- Schema layer is now a proper single source of truth: `InstrumentInfo` discriminated on `kind`, `CreateSeriesInstrumentResult` discriminated on `outcome`, request schema composed from `$InstrumentDetails`/`$ClientInstrumentDetails`/`$InstrumentLanguage`/`$ScalarInstrumentInternal`.
- Service tests are meaningful: duplicate-confirm-without-create, case-insensitive title uniqueness, series-only delete refusal, id-generation includes title, and the assertion that `instrumentRecordModel.count` is never called (`instruments.service.spec.ts:289`).
- `submitButtonLabel` correctly threaded as `LocalizedText` and resolved with `t()` at the leaf (`FormContent.tsx:51`) — fixes the "french" concern.
- Group-save 500 fix is minimal and well-scoped (`groups.service.ts:88-95`), with a test for stale ids.
- Discriminated-union consumers narrow on `kind` instead of optional-chaining (`InstrumentCard.tsx:51`, `instrument-repos/index.tsx:321`, `manage.tsx:901`).

thomasbeaudry:
Merged `origin/main` into this branch (`0369722a4`) — brings in the incomplete-series gateway fix, the series terminate function, and the synchronizer error-logging improvements. One conflict in `gateway.synchronizer.ts` was resolved by keeping main's restructured decrypt/parse/create error handling and re-applying the `seriesInstrumentId` metadata tagging from `fe5c7b6be` inside it. Typecheck, full lint (35/35), and api tests (106/106) all pass post-merge.

packages/schemas/src/instrument/instrument.base.ts
```diff
@@ -272,6 +297,38 @@ const $CreateInstrumentData = z.object({
...
 });
 
+/**
+ * The data a client submits to assemble a new series instrument from existing scalar instruments. The
+ * multilingual `details`/`clientDetails` are picked directly from the master instrument schemas (single
+ * source of truth); everything else about the stored instrument (id, content, kind, ...) is computed
+ * server-side. Exported as both a value (the schema) and a type so it can annotate a controller body
+ * without a dedicated DTO class.
+ */
+type $CreateSeriesInstrumentData = z.infer<typeof $CreateSeriesInstrumentData>;
```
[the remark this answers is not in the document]
thomasbeaudry: This compiles as written — `type $CreateSeriesInstrumentData` and `const $CreateSeriesInstrumentData` share a name via TypeScript declaration merging, and the export block at the bottom of the file exports that name in both its type and value meanings, so `import { $CreateSeriesInstrumentData }` works in either position (the monorepo `tsc` and CI lint-and-test pass on this code). The type/const same-name pattern is intentional here: the API controller imports it as a value so the zod schema itself becomes the body's `design:paramtypes` metadata, which libnest's ValidationPipe recognizes (`metatype instanceof z.ZodType`) — giving runtime validation without a DTO class, while the same identifier annotates the static type.

apps/web/src/routes/_app/group/manage.tsx
```diff
@@ -218,12 +318,283 @@ const InstrumentPreviewDialog = ({ item, onClose }: { item: InstrumentItem; onCl
...
+      return;
+    }
+    const result = await createMutation.mutateAsync(buildPayload(items));
+    if (result.outcome === 'duplicate') {
+      // The existing title comes back in its stored form (a plain string, or multilingual object).
+      setDuplicateOf(typeof result.existingTitle === 'string' ? result.existingTitle : t(result.existingTitle));
+      return;
+    }
+    onClose();
+  };
```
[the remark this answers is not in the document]
thomasbeaudry: Partly agreed. The intended UX is actually preserved on failure: `onClose()` sits after the `await`, so a rejection leaves the dialog open, and the mutation's react-query `onError` already raises the error notification. What does escape is the rejected promise itself from the `void handleCreate()` call site (console noise / unhandled-rejection event, no user-facing impact). A `try/catch` swallow at the handler level would silence that — worth adding as a small hardening pass alongside the two sibling handlers flagged below.

apps/web/src/routes/_app/group/manage.tsx
```diff
@@ -218,12 +318,283 @@ const InstrumentPreviewDialog = ({ item, onClose }: { item: InstrumentItem; onCl
...
+
+  const handleConfirmDuplicate = async () => {
+    const items = buildItems();
+    if (items.length < 2) {
+      return;
+    }
+    await createMutation.mutateAsync({ ...buildPayload(items), confirmDuplicate: true });
+    onClose();
+  };
```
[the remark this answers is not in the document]
thomasbeaudry: Same assessment as `handleCreate` above: the dialog already stays open on failure (state updates only run after the `await`) and the mutation's `onError` notification fires, so retry is possible — the genuine defect is limited to the rejection escaping the `void` call site uncaught. Same one-line `try/catch` hardening applies.

apps/web/src/routes/_app/group/manage.tsx
```diff
@@ -218,12 +318,283 @@ const InstrumentPreviewDialog = ({ item, onClose }: { item: InstrumentItem; onCl
...
+  onDeleted: (id: string) => void;
+}) => {
+  const { t } = useTranslation();
+  const deleteMutation = useDeleteInstrumentMutation();
+
+  const handleDelete = async () => {
+    await deleteMutation.mutateAsync({ id: item.id });
+    onDeleted(item.id);
+    onClose();
+  };
```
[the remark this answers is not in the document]
thomasbeaudry: Confirmed the success-path ordering is already correct: `onDeleted()`/local-state updates run only after `await deleteMutation.mutateAsync(...)` resolves, and `useDeleteInstrumentMutation.onError` shows the failure notification — so state is not corrupted on failure. As with the create handlers, the remaining issue is only the uncaught rejection from the `void handleDelete()` call site; same `try/catch` hardening applies to all three handlers.

apps/api/src/instruments/instruments.controller.ts
```diff
@@ -22,6 +29,23 @@ export class InstrumentsController {
     return this.instrumentsService.create(data);
   }
 
+  @ApiOperation({ summary: 'Create Series Instrument' })
+  @Post('series')
+  @RouteAccess({ action: 'create', subject: 'Instrument' })
+  createSeries(
```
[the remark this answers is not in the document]
thomasbeaudry: Accurate observation, and worth a maintainer decision. `create Instrument` for GROUP_MANAGER is required by design — creating series instruments on the fly is a core group-manager flow (and the upcoming bulk remote-assignment feature depends on it) — but it does also authorize `POST /v1/instruments` (arbitrary bundle upload), which the web UI never exposes to group managers. If managers should be restricted to series creation only, the cleanest split is a dedicated subject/action for the upload endpoint (e.g. require `manage` on `Instrument` there) while leaving `create` for `POST /instruments/series`. Flagging as a follow-up rather than changing it in this PR, since tightening it affects existing GROUP_MANAGER API behavior beyond the series feature.

apps/api/prisma/schema.prisma
```diff
@@ -189,7 +194,10 @@ model Instrument {
   bundle         String
   groups         Group[]            @relation(fields: [groupIds], references: [id])
   groupIds       String[]           @db.ObjectId
-  records        InstrumentRecord[]
+  records        InstrumentRecord[] @relation("InstrumentRecords")
```
joshunrau, line 198: Why is this `@relation` statement different than every other one in the schema? Please explain why this is here.

apps/api/prisma/schema.prisma
```diff
@@ -189,7 +194,10 @@ model Instrument {
...
+  seriesRecords  InstrumentRecord[] @relation("SeriesInstrumentRecords")
+  seriesGroup    Group?             @relation("OwnedSeriesInstruments", fields: [seriesGroupId], references: [id])
```
joshunrau, line 201: please justify why `@relation` has a string as the first argument in all cases

apps/api/src/auth/__tests__/ability.factory.test.ts
```diff
@@ -49,6 +49,21 @@ describe('AbilityFactory', () => {
...
+      basePermissionLevel: 'GROUP_MANAGER',
+      firstName: 'Test',
+      groups: [{ id: 'group-1' }],
+      id: 'user-1',
+      lastName: 'User',
+      username: 'manager-user'
+    } as any);
+
+    expect(ability.can('delete', subject('Instrument', { seriesGroupId: 'group-1' }) as any)).toBe(true);
+    expect(ability.can('delete', subject('Instrument', { seriesGroupId: 'group-2' }) as any)).toBe(false);
```
joshunrau, line 65: please also add a test with a null groupId. Group managers should not be able to delete series instruments with a null group id

apps/api/src/core/types.ts
```diff
@@ -2,4 +2,5 @@ import type { AppAbility } from '@/auth/auth.types';
 
 export type EntityOperationOptions = {
   ability?: AppAbility;
+  groupIds?: string[];
```
joshunrau: We should not be changing this. You should pass through `currentUser` like is the established pattern in the codebase.

apps/api/src/instruments/instruments.service.ts
```diff
@@ -102,13 +121,113 @@ export class InstrumentsService {
...
+  async deleteById(id: string, { ability }: EntityOperationOptions = {}): Promise<{ id: string }> {
+    const deleteAccess = accessibleQuery(ability, 'delete', 'Instrument');
+    const instrument = await this.instrumentModel.findFirst({
+      where: {
+        id,
+        OR: [
+          deleteAccess,
+          {
+            AND: [
+              { OR: [{ seriesGroupId: null }, { seriesGroupId: { isSet: false } }] },
```
joshunrau: why is the user allowed to delete instruments with seriesGroupId null?

joshunrau: this would allow any user to  delete any instrument wouldn't it?

apps/api/src/instruments/instruments.service.ts
```diff
@@ -178,57 +324,74 @@ export class InstrumentsService {
...
     return this.cryptoService.hash(`${name}-${edition}`);
   }
 
-  generateSeriesInstrumentId(instrument: SeriesInstrument) {
+  generateSeriesInstrumentId(instrument: SeriesInstrument, seriesGroupId?: string) {
     return this.cryptoService.hash(
-      getSeriesInstrumentItems(instrument.content)
-        .map(({ edition, name }) => `${name}-${edition}`)
-        .join('--')
+      JSON.stringify({
```
joshunrau, line 434: We already have series instruments in the database using the old hash format. This will break the check for duplicates on create. I am thinking we need something like a prefix `__V2__{}` and then we can infer it is your schema version. Open to anything.

joshunrau (changes requested):
much better, a few small changes requested

apps/api/src/instrument-records/instrument-records.service.ts
```diff
@@ -90,6 +98,17 @@ export class InstrumentRecordsService {
       );
     }
 
+    // When the record was collected through a series instrument, verify the reference points at an
+    // actual series before persisting it in the record's metadata.
+    if (seriesInstrumentId) {
+      const seriesInstrument = await this.instrumentsService.findById(seriesInstrumentId);
+      if (seriesInstrument.kind !== 'SERIES') {
```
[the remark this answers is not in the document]
thomasbeaudry: Addressed in 2e459ce. create() now forwards the caller options to both instrument lookups and scopes both lookups with groupIds: [groupId] when a record group is provided. I also added a regression test covering the ability and group scope passed to both lookups.

apps/api/src/instruments/instruments.controller.ts
```diff
@@ -22,31 +30,70 @@ export class InstrumentsController {
...
+    @Body() data: $CreateSeriesInstrumentData,
+    @CurrentUser('ability') ability: AppAbility
+  ): Promise<CreateSeriesInstrumentResult> {
+    return this.instrumentsService.createSeries(data, { ability });
+  }
+
+  @ApiOperation({ summary: 'Delete Instrument' })
```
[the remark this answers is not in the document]
thomasbeaudry: Addressed in 2e459ce. The Swagger summary now reads “Delete Series Instrument”.

apps/api/prisma/schema.prisma
thomasbeaudry: Addressed in 14a17a0. The string is Prisma’s relation name and must match on both sides of the relation. OwnedSeriesInstruments distinguishes the new Group-to-owned-series relation from the existing Group-to-Instrument access relation; without a name Prisma cannot disambiguate the two Group/Instrument relations. I added a schema comment for this as well.

apps/api/prisma/schema.prisma
thomasbeaudry: Addressed in 14a17a0. InstrumentRecord now has two distinct relations to Instrument: instrument/instrumentId is the administered scalar instrument, while seriesInstrument/seriesInstrumentId is the optional orchestrating series. Prisma requires explicit relation names to disambiguate those two model pairs, and InstrumentRecords matches the name on both sides. I also added a schema comment explaining this.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Addressed in 14a17a0. New series IDs now use the __V2__ prefix before the hash, making the new hash schema distinguishable from legacy unprefixed IDs. Legacy series remain included in semantic duplicate detection by comparing their exact ordered item sequence, and the regression test now explicitly exercises an unversioned legacy series while the ID-generation test verifies the __V2__ prefix.

apps/api/src/core/types.ts
thomasbeaudry: Addressed in 14a17a0. I reverted the groupIds addition to EntityOperationOptions. The instrument controllers now pass the full currentUser through to the service, and the service derives and validates the requested group scope from currentUser following the established pattern.

apps/api/src/auth/__tests__/ability.factory.test.ts
thomasbeaudry: Addressed in 14a17a0. The ability test now explicitly verifies that a group manager cannot delete an Instrument whose seriesGroupId is null. I also removed the service-level null-owner fallback, so deletion is governed solely by the caller’s delete ability.

apps/api/src/instruments/instruments.service.ts
thomasbeaudry: Addressed in 14a17a0. I removed the fallback that made null/unset seriesGroupId instruments deletable. deleteById now queries only through accessibleQuery using currentUser.ability, so group managers can delete only series owned by one of their groups; null-owned and other-group series are rejected.

apps/api/src/auth/ability.factory.ts
```diff
@@ -27,6 +27,12 @@ export class AbilityFactory {
         ability.can('manage', 'Assignment', { groupId: { in: groupIds } });
         ability.can('manage', 'Group', { id: { in: groupIds } });
         ability.can('read', 'Instrument');
+        // Group managers may assemble series instruments on the fly and delete ones they created. A
+        // series is a bundle of other instruments, not a shared platform asset. Read and delete access
+        // to generated series is restricted to the group that owns it; shared and legacy instruments
+        // have no owning group and remain readable by all groups.
+        ability.can('create', 'Instrument');
```
gdevenyi, line 34: 🔴 Still open from the earlier pass and worth a decision before this merges.

On `main`, `GROUP_MANAGER` only had `read` on `Instrument`. This PR adds an **unscoped** `create`, which `createSeries` needs — but the regular scalar-upload route `POST /v1/instruments` (`instruments.controller.ts:27`) is gated by the exact same `action: 'create', subject: 'Instrument'`. Net effect: group managers can now upload arbitrary scalar instrument bundles via `POST /v1/instruments` (executed server-side in the virtualization sandbox) — previously admin-only. That looks incidental to the series feature rather than part of the design.

If series-only is intended, the cleanest split is to tighten the upload route (e.g. gate `POST /v1/instruments` on `manage`/admin) and leave `create` for `POST /instruments/series`, or introduce a dedicated subject for series creation. Copilot flagged the same thing at `87ce6b5` and it's currently marked follow-up — re-raising because it's a privilege expansion rather than a styling point.

apps/web/src/hooks/useDeleteInstrumentMutation.ts
```diff
@@ -0,0 +1,33 @@
...
+    },
+    onSuccess() {
+      addNotification({ type: 'success' });
+      // The server has already detached the instrument from every group; callers that hold a stale copy
+      // of the group's accessible ids (e.g. the manage page) reconcile it locally via the mutation result.
+      void queryClient.invalidateQueries({ queryKey: ['instrument-info'] });
+    },
+    // Callers catch mutateAsync failures and this hook presents the API message as a notification.
+    // Do not also rethrow the same handled error into the route error boundary.
+    throwOnError: false
```
gdevenyi, line 31: 🟡 Minor / question: for `useMutation`, `throwOnError` controls whether errors bubble to the error boundary when invoking via `mutate` — `mutateAsync`'s returned promise still rejects on failure regardless of this flag. Since `deleteSeriesInstrument` already wraps the `mutateAsync` call in `try/catch`, `throwOnError: false` isn't doing additional work at this call site. Harmless, but the comment reads as if it changes `mutateAsync`'s rejection behavior. Confirming whether this was meant to guard a different (non-`mutateAsync`) call site, or whether the try/catch alone is sufficient here.

apps/web/src/routes/_app/group/delete-series-instrument.ts
```diff
@@ -0,0 +1,22 @@
...
+export async function deleteSeriesInstrument({
+  deleteInstrument,
+  id,
+  onClose,
+  onDeleted
+}: DeleteSeriesInstrumentOptions): Promise<void> {
+  try {
+    await deleteInstrument({ id });
+    onDeleted(id);
+  } catch {
```
gdevenyi, line 17: 🟡 Question on failure UX: `finally { onClose() }` dismisses the dialog on failure too, whereas `CreateSeriesInstrumentDialog` deliberately stays open on failure so the user can retry. For delete that's arguably fine (no form state to preserve, retry happens from the list), but the empty `catch` + always-close means a failed delete silently dismisses the dialog and leaves only the toast. Intentional asymmetry, or should the dialog stay open on failure for parity with create?

gdevenyi (commented):
**Follow-up review on `14a17a0`** (since my last pass at `c2478b9`).

The three commits since then cleanly close out the open items from both my earlier review and @joshunrau's "changes requested":

- **Group-scoped ownership** — series now carry `seriesGroupId` + a named `seriesGroup` relation; the delete ability is `{ seriesGroupId: { in: groupIds } }`, and the ability test asserts a manager can't delete a series with `seriesGroupId: null` or one owned by another group. Resolves the "any user can delete any instrument" concern.
- **`currentUser` pass-through** — controllers hand the full `RequestUser` to the service (the established pattern); the `core/types.ts` change was reverted.
- **`create()` auth scope** — `InstrumentRecordsService.create` forwards `options` + `groupIds` to both `findById` lookups, so a `seriesInstrumentId` from another group can no longer be attached. Regression test added.
- **Legacy vs V2 series ids** — new ids get the `__V2__` prefix; legacy (unprefixed) series are still picked up by duplicate detection via ordered-sequence comparison. Both paths tested.
- **Reproducibility guard** — `deleteById` now refuses a series once any record references it (counting both `instrumentId` and `seriesInstrumentId`), matching the concern raised in the PR description.
- **Schema disambiguation** — named `@relation` on both sides with explanatory comments.
- **Swagger** — "Delete Series Instrument".

Two things are still open from the prior pass:

1. 🔴 **The `GROUP_MANAGER` `create` over-grant on `POST /v1/instruments`** — inline comment on `ability.factory.ts:34`. This is the one item I'd want explicitly signed off before merge.
2. 🟢 **`pnpm-lock.yaml`** — still ~8.6k lines of formatting-only churn with no `package.json` changes anywhere in the PR. Regenerating against the repo's pinned pnpm would keep the diff reviewable and avoid future merge conflicts.

Two minor nits inline (`useDeleteInstrumentMutation` `throwOnError`, `delete-series-instrument` failure UX).

Everything else from the last pass looks resolved. The group-ownership rework is a meaningful improvement — the model is much cleaner than the `sourceRepoId: null` scoping it replaces.

apps/web/src/hooks/useDeleteInstrumentMutation.ts
thomasbeaudry: Confirmed that mutateAsync rejects regardless of throwOnError. The local false override is still required here because the app globally configures mutations with throwOnError: true in services/react-query.ts: after the caller catches the promise rejection, the mutation error state otherwise throws during the hook’s next render and reaches the route boundary. Removing the override reproduces that failure in instrument-group-scope.test.tsx. I kept the override in a55d8b7 and corrected the comment to describe the global-default override and mutateAsync behavior precisely.

apps/api/src/auth/ability.factory.ts
thomasbeaudry: Addressed in a55d8b7. The regular POST /v1/instruments bundle-upload route now requires manage on Instrument, which remains admin-only for the existing permission model. POST /v1/instruments/series still uses create, so group managers can create group-owned series without gaining arbitrary scalar bundle upload access. The ability test now explicitly verifies create is allowed while manage is denied for a group manager.

apps/web/src/routes/_app/group/delete-series-instrument.ts
thomasbeaudry: This asymmetry is intentional. For delete there is no form state to preserve, and the requested UX for an already-used/protected series is to show the API notification and dismiss the confirmation popup rather than leave it stuck open. The rejected-delete test verifies that the dialog closes without removing the instrument locally. I clarified that intent in the catch comment in a55d8b7.

PR #1397, opened 2026-07-14
feat(web): group switcher, tutorial hide, WebP + 2MB login images

thomasbeaudry:
## Summary

Tweaks the login walkthrough, login-page branding, and the dashboard/group selector.

- **Group switcher** — Adds a `Current Group` dropdown, letting users switch group from any authenticated page rather than only the dashboard. Its placement is a user preference (**Settings → Group switcher position**): `sidebar` (default) or `topbar`. The per-page group dropdown is removed from the dashboard, and the dashboard title reverts to plain "Dashboard".
- **Walkthrough "Don't show again"** — Adds a button to permanently dismiss the login tutorial, and carries the persisted flag across store-version bumps so the choice survives ODC upgrades. The Tutorial menu item still relaunches it on demand.
- **Login image** — Raises the upload limit from 1 MB to 2 MB, renames the section from "Logo" to "Login Image", and renders uploaded images over a checkerboard in the editor so transparency is visible.
- **WebP support** — The file input's `accept` already listed `image/webp`, but nothing validated the selected file and all user-facing copy said "SVG, PNG, JPEG". The accepted types are now a single `ACCEPTED_LOGO_MIME_TYPES` constant driving both the `accept` hint and a real check in `handleLogoFile` (the OS picker can be overridden with "All Files"), and the editor copy names WebP.
- **WebP re-encoding** — The uploaded image is stored verbatim as a base64 data URI on the `SetupState` document, so a 2 MB PNG was carried as ~2.7 MB on every read. Raster uploads are now re-encoded to WebP in the editor via the browser's own canvas encoder — **no new dependency**, so no impact on image size (server-side conversion would have meant a native codec like `sharp`). Measured on generated logo rasters: **61-69% smaller**. SVG and existing WebP pass through untouched; the result is discarded if it comes out no smaller or if the browser has no WebP encoder (`toDataURL` silently falls back to PNG rather than throwing).
- **Logo source hardening** — Both `customLogoSrc` and `customLogoUrl` render into ``, but neither was gated server-side: `customLogoSrc` accepted any string, and `z.string().url()` accepts a `javascript:` URL, so a crafted request could store either. `customLogoSrc` is now restricted to a base64 image data URI (or an http(s) URL, for logos saved before `customLogoUrl` existed) and `customLogoUrl` to an http(s) scheme — mirroring the gate already applied to resource-link hrefs. Covered by `packages/schemas/src/setup/setup.test.ts`.

Note: the schema deliberately allows any `image/*` subtype rather than only the four the editor offers, because `$BrandingConfig` also gates reads (`setup.service.ts`), where a rejected value drops the *entire* branding config. The narrower list is enforced client-side.

## Status / notes
- `turbo run lint` passes across all 29 non-database packages; `vitest` passes (258 tests).
- Open question: whether to disable the group switcher during an in-progress session (switching group clears the session — existing behavior, now reachable from every page).


gdevenyi:
5MB is huge. Pretty sure larger than the rest of the page.

Is this png? Its not really a modern format, should be using JPG

thomasbeaudry:
The image can be:  SVG, PNG, JPEG    I was thinking of bumping it to 2MB, but i figured 5MB is still small.  Note that it's not really a logo, maybe i can come up a better name, it's for the image that gets show on the left pane of the login page (in the attached pic, it's the mouseODC image), which is customizable so in theory it could span a large area, so a larger size makes sense.

image

gdevenyi:
2MB is OK but 5MB is extreme.

SVG and JPG both would be relatively tiny versus a lossless PNG

thomasbeaudry:
Latest push:

- Renamed "Logo" section to "Login Image" / "Image de connexion" in the branding editor
- Reduced image upload limit from 5 MB to 2 MB (schema + UI strings)
- Darkened sidebar group switcher from `bg-primary` to `bg-sky-700` so it blends better with the slate sidebar

thomasbeaudry:
note that converseion of login image to webp is done in the browser already so we don't need sharp

gdevenyi (commented):
## Review

Overall this is a solid PR: the schema hardening for `customLogoSrc`/`customLogoUrl` with tests is the right call, the WebP re-encode approach (browser canvas, no new dependency, discard-if-not-smaller) is well designed, and the `image.ts` module is clearly documented. Inline comments below; the two things I'd consider blocking are the **group-switch-clears-session** behavior (now reachable from every page) and the **triplicated switcher UI**.

### On the open question in the description
Yes — guard it. `changeGroup` silently sets `currentSession: null`, and this PR moves that action from the dashboard to every authenticated page. A user mid-session who flicks the dropdown loses their session context with no warning. The Sidebar already has exactly the right pattern: the end-session confirm `Dialog`. Reuse it: when `currentSession !== null`, confirm before switching (see inline comment).

### On the existing Copilot comments
- **`uploaderEnabled` never syncs** (settings.tsx): moot — `useSetupStateQuery` is a `useSuspenseQuery`, so `data` is always defined at first render and the lazy `useState` initializer is safe. No change needed.
- **`localStorage` can throw** (auth.slice.ts ×2): the literal try/catch suggestion treats the symptom. The store already persists preferences through zustand `persist` (which wraps the same `localStorage`); moving the preferred-group id into the preferences slice removes both raw `localStorage` calls and both comments (see inline comment).
- **`HTTP_URL_SCHEME_PATTERN` accepts `http://` alone**: valid, and there's a one-line fix — `RESOURCE_LINK_URL_PATTERN` is defined 15 lines up in the same file (see inline comment).
- **`NavGroup` doesn't auto-expand on route change** and **`as GroupSwitcherPosition` cast**: both valid, endorsed inline.

### Minor, no inline comment
- `store/index.ts` `migrate: (persistedState) => persistedState as AppStore` — fine today, but it's a blanket "keep everything" that will silently no-op the day a version bump actually needs to reshape persisted state. The comment explains the intent well; just remember this function is where that future work goes.
- The single-group case renders a non-interactive `div` styled like the select trigger (`bg-primary`, same width). Looks clickable but isn't; consider muted styling.
- Nice touches: `key={currentGroup?.id}` remount fix in `group/manage.tsx`, checkerboard transparency preview, and the `accept`-hint-plus-real-validation pairing driven by one constant.

apps/web/src/components/GroupSwitcher/GroupSwitcher.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <div
+          className="bg-primary text-primary-foreground flex h-9 w-[180px] items-center justify-center rounded-md text-sm font-semibold"
+          data-testid="group-switcher-select"
+        >
+          {currentGroup.name}
+        </div>
+      ) : (
+        <Select
+          value={currentGroup.id}
+          onValueChange={(id) => changeGroup(currentUser.groups.find((group) => group.id === id)!)}
```
gdevenyi: **Session loss is now one misclick away.** `changeGroup` clears `currentSession` with no warning, and this control now lives on every authenticated page (in all three variants — here, Sidebar, Navbar). Suggest: when `currentSession !== null`, open a confirm dialog before calling `changeGroup` — the Sidebar's end-session `Dialog` is the existing pattern to reuse. That also answers the open question in the PR description: don't disable the switcher mid-session (hidden state is confusing), confirm instead.

apps/web/src/components/Sidebar/Sidebar.tsx
```diff
@@ -34,18 +44,50 @@ export const Sidebar = () => {
...
       </div>
       <hr className="my-2 h-[1px] border-none bg-slate-700" />
+      {showSidebarGroupSwitcher && (
+        <div className="mb-2">
+          {isSingleGroup ? (
+            <div className="flex h-9 w-full items-center justify-center rounded-md bg-sky-700 text-sm font-semibold text-slate-100">
+              {currentGroup.name}
+            </div>
+          ) : (
+            <Select
```
gdevenyi: **Third copy of the same UI.** This single-group-badge + `Select` block is duplicated nearly verbatim in `GroupSwitcher.tsx`, here, and `Navbar.tsx` (the only deltas are `bg-primary` vs `bg-sky-700` and the wrapper). A rename or behavior change (e.g. the session-loss confirm) now has to land in three places. Give `GroupSwitcher` a `triggerClassName`/variant prop and render it in all three spots — this file and Navbar each shrink by ~25 lines.

apps/web/src/store/slices/auth.slice.ts
```diff
@@ -18,22 +18,27 @@ const parseAccessToken = (accessToken: string) => {
...
+const PREFERRED_GROUP_KEY = 'odc-preferred-group-id';
+
 export const createAuthSlice: SliceCreator<AuthSlice> = (set) => {
   const accessToken = window.__PLAYWRIGHT_ACCESS_TOKEN__ ?? null;
   const initialState = accessToken ? parseAccessToken(accessToken) : null;
 
   return {
     accessToken,
     changeGroup: (group) => {
+      localStorage.setItem(PREFERRED_GROUP_KEY, group.id);
```
gdevenyi: Two issues with the raw `localStorage` approach:

1. **It bypasses the persistence pattern this same PR introduces.** `groupSwitcherPosition` goes through the preferences slice + `partialize`; the preferred group id is the same kind of per-browser preference and belongs there too (`preferredGroupId` in `PreferencesSlice`, added to `partialize`, read in `login`). That also resolves both Copilot try/catch comments for free — zustand's `persist` already guards storage access.
2. **The key isn't scoped per user.** On a shared workstation, user A switches to group X, logs out; user B logs in and — if they're also a member of X — silently lands in X instead of their token's default group. In a clinical multi-group setup that's a plausible mis-filing vector. Key the stored value by username (e.g. `{ [username]: groupId }`).

apps/web/src/hooks/useNavItems.ts
```diff
@@ -22,10 +23,11 @@ import { useAppStore } from '@/store';
 import { useSetupStateQuery } from './useSetupStateQuery';
 
 export type NavItem = {
+  children?: NavItem[];
   disabled?: boolean;
   icon: React.ComponentType<Omit<React.SVGProps<SVGSVGElement>, 'ref'>>;
   label: string;
-  url: string;
+  url?: string;
```
gdevenyi, line 30: Making `url` optional forces `url!` non-null assertions at four call sites (Sidebar, Navbar ×2, NavGroup) — against the repo convention of avoiding casts. A discriminated union captures the actual shape (leaf has `url`, group has `children`) and deletes every assertion:

```ts
type NavItemBase = {
  disabled?: boolean;
  icon: React.ComponentType<Omit<React.SVGProps<SVGSVGElement>, 'ref'>>;
  label: string;
};
export type NavItem = NavItemBase &
  ({ children?: never; url: string } | { children: NavItem[]; url?: never });
```

`props.children ? <NavGroup/> : <NavButton/>` then narrows correctly on both branches, and `NavButton` can go back to plain `NavItem &` without the `Omit`.

apps/web/src/routes/_app/admin/settings.tsx
```diff
@@ -1,61 +1,123 @@
...
+            <Card.Title className="text-lg font-bold">{t({ en: 'Preferences', fr: 'Préférences' })}</Card.Title>
+          </Card.Header>
+          <Card.Content>
+            <div className="flex items-center justify-between gap-4">
+              <p className="text-sm font-medium">
+                {t({ en: 'Group Switcher Position', fr: 'Position du sélecteur de groupe' })}
+              </p>
+              <Select
+                value={groupSwitcherPosition}
+                onValueChange={(value) => setGroupSwitcherPosition(value as GroupSwitcherPosition)}
```
gdevenyi, line 131: Three things on this control:

1. **Placement**: this is a per-browser user preference, but it lives on a page whose nav entry requires `ability.can('manage', 'all')` — a non-admin clinician can never find or change it. If it's meant to be per-user, it belongs somewhere every user can reach (e.g. a user preferences section); if it's meant to be instance-wide, it should live in `SetupState`, not `localStorage`.
2. **Inconsistent save semantics**: this Select applies instantly, while the uploader toggle above requires the Save button. On one card that reads as "Save applies to both" — either make the Select part of Save, or move it out of this card.
3. Copilot's point on `value as GroupSwitcherPosition` stands; typing the handler avoids the cast: `onValueChange={(value: GroupSwitcherPosition) => setGroupSwitcherPosition(value)}` (the values come from the two `Select.Item`s below, so this is safe).

packages/schemas/src/setup/setup.ts
```diff
@@ -53,6 +53,22 @@ const $BrandingText = z.object({
...
+ * `customLogoSrc` may also hold an http(s) URL for logos saved before
+ * `customLogoUrl` existed (see `logoSource`).
+ *
+ * Any `image/*` subtype is allowed rather than just the four the editor offers:
+ * this schema also gates reads, where a rejected value drops the whole branding
+ * config, so it must stay permissive enough for images saved by earlier versions.
+ * The narrower list the editor accepts is enforced client-side.
+ */
+const LOGO_DATA_URI_PATTERN = /^data:image\/[a-zA-Z0-9.+-]+;base64,[A-Za-z0-9+/]+={0,2}$/;
+const HTTP_URL_SCHEME_PATTERN = /^https?:\/\//;
```
gdevenyi, line 75: Copilot's point here is valid and the fix is already in this file: `RESOURCE_LINK_URL_PATTERN` (15 lines up) is the vetted http(s)-URL shape used for resource-link hrefs, and the legacy `customLogoSrc` values this branch exists for were validated against the same editor `URL_PATTERN`. Using it for both fields drops `HTTP_URL_SCHEME_PATTERN` entirely, and makes the `.url()` call on `customLogoUrl` redundant (the pattern is stricter):

```ts
customLogoSrc: z.string().max(3_000_000)
  .refine((value) => LOGO_DATA_URI_PATTERN.test(value) || RESOURCE_LINK_URL_PATTERN.test(value), ...)
customLogoUrl: z.string().max(2000).regex(RESOURCE_LINK_URL_PATTERN, 'Must be an http(s) URL').nullish(),
```

apps/web/src/components/NavGroup/NavGroup.tsx
```diff
@@ -0,0 +1,67 @@
...
+  childClassName,
+  className,
+  icon: Icon,
+  items,
+  label,
+  onNavigate
+}: NavGroupProps) => {
+  const location = useLocation();
+  const containsActive = items.some((item) => item.url === location.pathname);
+  const [isOpen, setIsOpen] = React.useState(containsActive);
```
gdevenyi, line 32: Copilot's comment is right: `containsActive` only seeds the initial state, so navigating to a child route by any other means (deep link, redirect, walkthrough) leaves the group collapsed while its child is active. One effect fixes it without fighting the user's manual collapse:

```ts
React.useEffect(() => {
  if (containsActive) setIsOpen(true);
}, [containsActive]);
```

apps/web/src/components/LoginPageEditor/hooks.ts
```diff
@@ -185,17 +187,39 @@ export const useBrandingForm = () => {
...
+        message: t({ en: 'The selected image is larger than 2 MB.', fr: "L'image sélectionnée dépasse 2 Mo." }),
         title: t({ en: 'File too large', fr: 'Fichier trop volumineux' }),
         type: 'error'
       });
       return;
     }
-    const reader = new FileReader();
-    reader.onload = () => update('customLogoSrc', reader.result as string);
-    reader.readAsDataURL(file);
+    readLogoAsWebpDataUrl(file)
```
gdevenyi, line 210: Small race: if the user picks a large PNG (slow `createImageBitmap` + encode) and then immediately picks a second file, the first promise can resolve last and overwrite the second selection. A generation counter closes it:

```ts
const logoRequestRef = useRef(0);
// in handleLogoFile:
const requestId = ++logoRequestRef.current;
readLogoAsWebpDataUrl(file)
  .then((dataUrl) => {
    if (logoRequestRef.current === requestId) update('customLogoSrc', dataUrl);
  })
```

Low likelihood, but the failure mode (admin saves the wrong image believing they replaced it) is confusing enough to be worth three lines.

joshunrau (approved):

PR #1398, opened 2026-07-14
fix(api): explicitly inject S3Client into StorageService

thomasbeaudry:
## Summary

`StorageService` never actually receives its `S3Client`, so **the API crash-loops on startup whenever `STORAGE_ENABLED=true`**. This makes any storage-enabled deployment (self-hosted rustfs/MinIO, file instruments, etc.) unusable. The fix is a one-line DI annotation.

## Symptom

With `STORAGE_ENABLED=true`, the `api` container restarts continuously. Logs show:

```
ServiceUnavailableException: File storage is not configured
    at StorageService.requireStorage (/app/src/storage/storage.service.ts:97:19)
    at StorageService.onModuleInit (/app/src/storage/storage.service.ts:78:37)
  response: { message: 'File storage is not configured', error: 'Service Unavailable', statusCode: 503 }
```

The environment is fully configured — `STORAGE_ENABLED`, `STORAGE_ACCESS_KEY`, `STORAGE_SECRET_KEY`, `STORAGE_BUCKET`, and `STORAGE_ENDPOINT` are all present in the container — yet `requireStorage()` reports storage as unconfigured.

## Root cause

`requireStorage()` throws when `!this.enabled || !this.s3 || !this.bucket`. Since `onModuleInit()` only reaches `requireStorage()` after passing its own `if (!this.enabled ...) return` guard, `enabled` is truthy and `bucket` is set — so **`this.s3` is `null`**.

`this.s3` is the injected `S3Client`. The constructor parameter is:

```ts
constructor(
  private readonly configService: ConfigService,
  @Optional() private readonly s3: null | S3Client
) { ... }
```

The parameter type is the union **`null | S3Client`**. TypeScript emits the reflected `design:paramtypes` metadata for a `null | X` union as **`Object`**, not `X`. Confirmed in the compiled bundle:

```js
StorageService = _ts_decorate([
  Injectable(),
  _ts_param(1, Optional()),
  _ts_metadata("design:paramtypes", [
    ConfigService,
    Object            // <-- s3 param: erased to Object, not S3Client
  ])
]);
```

Nest therefore has no usable token to resolve the parameter by type. Because the parameter is `@Optional()`, Nest does not error — it **silently injects `null`**. The `S3Client` provider registered in `StorageModule` is never wired into `StorageService`, so `this.s3` is *always* `null`, independent of configuration. The factory that builds the real client (gated on `STORAGE_ENABLED`) runs fine; its result is just never delivered.

## Fix

Add an explicit `@Inject(S3Client)` so Nest resolves the parameter by token rather than by the union-erased design type:

```ts
@Optional() @Inject(S3Client) private readonly s3: null | S3Client
```

The nullable type is retained on purpose: the provider legitimately returns `null` when `STORAGE_ENABLED` is false, and `@Optional()` remains correct for that case.

## Verification

Built the production `api` image from this branch and ran the full docker-compose stack with `STORAGE_ENABLED=true` against a rustfs backend:

- **Before:** `api` in `Restarting` state, looping on `ServiceUnavailableException: File storage is not configured`.
- **After:** `api` stays `Up`; no storage error; `GET /api/v1/setup` returns `{"release":{"type":"production","version":"2.1.2"}, ...}` and the bucket is created/verified on init.

### DIFF
--- a/apps/api/src/storage/storage.service.ts
+++ b/apps/api/src/storage/storage.service.ts
@@ -7,7 +7,7 @@ import {
 } from '@aws-sdk/client-s3';
 import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
 import { ConfigService } from '@douglasneuroinformatics/libnest';
-import { Injectable, Optional, ServiceUnavailableException } from '@nestjs/common';
+import { Inject, Injectable, Optional, ServiceUnavailableException } from '@nestjs/common';
 import type { OnModuleInit } from '@nestjs/common';
 import type { $FileLocation, $PresignedUrlInfo } from '@opendatacapture/schemas/storage';
 
@@ -28,7 +28,7 @@ export class StorageService implements OnModuleInit {
 
   constructor(
     private readonly configService: ConfigService,
-    @Optional() private readonly s3: null | S3Client
+    @Optional() @Inject(S3Client) private readonly s3: null | S3Client
   ) {
     this.enabled = !!this.configService.get('STORAGE_ENABLED');
     this.bucket = this.configService.get('STORAGE_BUCKET');


apps/api/src/storage/storage.service.ts
```diff
@@ -28,7 +28,7 @@ export class StorageService implements OnModuleInit {
 
   constructor(
     private readonly configService: ConfigService,
-    @Optional() private readonly s3: null | S3Client
+    @Optional() @Inject(S3Client) private readonly s3: null | S3Client
```
gdevenyi, line 31: This is the right call: `@Inject(S3Client)` pins the token explicitly so the `null | S3Client` union (which erases to `Object` in `design:paramtypes`) can't shadow resolution. Keeping `@Optional()` + the nullable type is correct since the provider returns `null` when storage is disabled.

One non-blocking suggestion: add a `StorageService`/`StorageModule` test that asserts the `S3Client` actually gets injected when `STORAGE_ENABLED=true`. Today there's no coverage at this boundary (only a mock in `instrument-records.service.spec.ts`), so this regression was invisible to the suite.

gdevenyi (approved):
## Verdict

Approving. The diagnosis is correct and the fix is minimal and idiomatic.

## Root cause — confirmed

`StorageModule` registers the client under the class token:

```ts
{ inject: [ConfigService], provide: S3Client, useFactory: ... }
```

`StorageService` had the parameter typed as `null | S3Client` with only `@Optional()`. Under `emitDecoratorMetadata`, a `null | S3Client` union collapses to `Object` in the emitted `design:paramtypes`, so Nest had no resolvable type token. Because the param is `@Optional()`, Nest silently injected `null` instead of throwing — leaving `this.s3` permanently `null` regardless of config. When `STORAGE_ENABLED=true`, `onModuleInit()` → `requireStorage()` throws `ServiceUnavailableException` and the API crash-loops. This matches the described symptom exactly.

## Fix — correct

`@Optional() @Inject(S3Client)` resolves the parameter by explicit token regardless of the union-erased metadata. Keeping the nullable type is right: the provider legitimately returns `null` when storage is disabled, and `@Optional()` correctly covers that case (and prevents a hard failure if the provider is ever absent).

No secondary concerns:
- Token consistency — `@Inject(S3Client)` matches the `provide: S3Client` registration. ✓
- No behavioral change for the disabled path (`this.s3` stays `null`, `requireStorage()` still guards with `!this.s3`). ✓
- Import ordering follows the existing `@nestjs/common` grouping. ✓

## Suggested follow-up (non-blocking)

There's no `StorageService` spec — it's only mocked in `instrument-records.service.spec.ts` — which is exactly why a DI wiring break at the module boundary went undetected. A small test that compiles `StorageModule` with `STORAGE_ENABLED=true` (and a stub `S3Client`/env) and asserts `service.isEnabled` / that `onModuleInit()` doesn't throw would lock this in and catch future provider/token drift. Happy to file a separate issue/PR for that so this fix stays scoped.

Nice root-cause writeup, by the way.

apps/api/src/storage/storage.service.ts
gdevenyi, line 31: Right call: `@Inject(S3Client)` pins the token explicitly so the `null | S3Client` union (which erases to `Object` in `design:paramtypes`) can't shadow resolution. Keeping `@Optional()` + the nullable type is correct since the provider returns `null` when storage is disabled.

Non-blocking suggestion: add a `StorageService`/`StorageModule` test asserting `S3Client` is actually injected when `STORAGE_ENABLED=true`. There's currently no coverage at this boundary (only a mock in `instrument-records.service.spec.ts`), so this regression was invisible to the suite.

thomasbeaudry: Agreed — the mock-only coverage at this boundary is exactly what let the bug slip through. I'll put up a follow-up PR with a `StorageModule` compilation test (stub S3Client + `STORAGE_ENABLED=true` env) that asserts the real provider wires through. Keeping it separate so this fix can land as-is.

thomasbeaudry: (Same thread — will address in the follow-up PR mentioned above.)

joshunrau (approved):

PR #1404, opened 2026-07-16
fix pending tag issue which causes old data not to be visible

david-roper:
claude fable 5


joshunrau (approved):

gdevenyi:
@david-roper This should've had a complete PR description.

PR #1405, opened 2026-07-17
fix: prevent DataTable pagination reset on row click

thomasbeaudry:
## Summary
- Upgraded the shared `@douglasneuroinformatics/libui` catalog to v6.9.1.
- Passes a stable highlight-color CSS variable through each DataTable `rootStyle` prop.
- Uses memoized DataTables plus context-driven row markers, so highlighting updates without rerendering DataTable and resetting pagination.
- Applies the fix to datahub subjects, admin users, admin groups, and instrument repositories.
- Adds a regression test covering page 2 row selection and highlight state.

## Validation
- [x] Regression test: `DataTableRowHighlight.test.tsx`
- [x] `pnpm lint`
- [x] `pnpm test` (38 files, 254 passed, 1 skipped)
- [ ] Manual authenticated route QA (local app reached login; credentials unavailable)

🤖 Continued with Codex after Claude Code session ended.


joshunrau (changes requested):
please fix all linting issues

apps/web/src/routes/_app/admin/groups/index.tsx
```diff
@@ -147,6 +132,24 @@ const RouteComponent = () => {
...
+    <Sheet open={Boolean(selectedGroup)} onOpenChange={() => setSelectedGroup(null)}>
+      <PageHeader>
+        <Heading className="text-center" variant="h2">
+          {t({
+            en: 'Manage Groups',
+            fr: 'Gérer les groupes'
+          })}
+        </Heading>
+      </PageHeader>
+      {highlightedRowIndex >= 0 && (
```
joshunrau: Okay, can you please update libui to [v6.9.1](https://github.com/DouglasNeuroInformatics/libui/tree/v6.9.1) and then pass this into the rootStyle prop of the datatable

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
```diff
@@ -137,40 +137,17 @@ const RouteComponent = () => {
...
-      <PageHeader>
-        <Heading className="text-center" variant="h2">
-          {t({
-            en: 'Instrument Repositories',
-            fr: "Dépôts d'instruments"
-          })}
-        </Heading>
-      </PageHeader>
+  const handleDeleteClickRef = useRef(handleDeleteClick);
+  handleDeleteClickRef.current = handleDeleteClick;
```
joshunrau: what is the point of this?, you just set it?

apps/web/src/routes/_app/admin/users/index.tsx
```diff
@@ -435,6 +420,24 @@ const RouteComponent = () => {
...
+      <PageHeader>
+        <Heading className="text-center" variant="h2">
+          {t({
+            en: 'Manage Users',
+            fr: 'Gérer les utilisateurs'
+          })}
+        </Heading>
+      </PageHeader>
+      {highlightedRowIndex >= 0 && (
+        <style>{`[data-testid="data-table-body"] > [id="${highlightedRowIndex}"] { background-color: var(--color-accent) }`}</style>
```
joshunrau: pass rootStyle

joshunrau (changes requested):

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
thomasbeaudry: Replaced the manual ref with libui useEventCallback. It keeps the memoized DataTable handler stable while reading current group assignments and translations.

apps/web/src/routes/_app/admin/users/index.tsx
thomasbeaudry: Done for users, groups, instrument repositories, and datahub subjects. Added a regression test covering selection on page 2.

apps/web/src/routes/_app/admin/groups/index.tsx
thomasbeaudry: Updated the libui catalog to ^6.9.1. All four affected DataTables now pass the highlight color through rootStyle; the app CSS consumes that inherited variable on the selected row.

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
```diff
@@ -233,11 +222,34 @@ const RouteComponent = () => {
...
+    >
+      <PageHeader>
+        <Heading className="text-center" variant="h2">
+          {t({
+            en: 'Instrument Repositories',
+            fr: "Dépôts d'instruments"
+          })}
+        </Heading>
+      </PageHeader>
+      <DataTableRowHighlightProvider rowId={highlightedRowId}>{dataTable}</DataTableRowHighlightProvider>
```
joshunrau, line 255: why do we have this component when it is only used once?

joshunrau (changes requested):

apps/web/src/routes/_app/admin/instrument-repos/index.tsx
thomasbeaudry: It's shared across all four routes that highlight rows: groups, users, instrument-repos, and datahub. The context + marker pattern lets highlighted rows update without re-rendering the memoized DataTable (which would reset pagination).

PR #1420, opened 2026-07-20
perf(datahub): resolve record usernames server-side

### LINKED ISSUE #1410: Performance: datahub downloads every session in the group to attach usernames, then joins O(n*m) in the browser

To display a username column alongside a subject's instrument records, the datahub fetches **every session in the current group** and joins it to the records in the browser with a linear scan per record.

Server side, `GET /v1/sessions` maps to `SessionsService.findAllIncludeUsernames` (`apps/api/src/sessions/sessions.service.ts:99`):

```ts
const sessionsWithUsers = await this.sessionModel.findMany({
  include: {
    subject: true,                        // full subject document, per session
    user: { select: { username: true } }
  },
  where: { AND: [accessibleQuery(ability, 'read', 'Session'), { groupId }] }
});
```

There is no pagination, no `subjectId` filter and no limit — the only filter is `groupId`. Every session ever created in the group is returned, each carrying an embedded copy of its subject.

Client side, `useInstrumentVisualization` (`apps/web/src/hooks/useInstrumentVisualization.ts:61`) requests it with only `groupId`:

```ts
const sessionsUsernameQuery = useFindSessionQuery({
  enabled: instrumentId !== null,
  params: { groupId: currentGroup?.id }
});
```

and then joins with a nested linear search (`:226-240`):

```ts
const records = recordsQuery.data.map((record) => {
  const usersSession = sessions.find((s) => s.id === record.sessionId);   // O(sessions) per record
  const username = usersSession?.user?.username ?? 'N/A';
  // ...
});
```

So viewing one subject's records for one instrument — typically a few dozen rows — downloads the session history of the entire group and performs `records × sessions` comparisons to attach a username to each row. This hook backs both `/datahub/$subjectId/table` and `/datahub/$subjectId/graph`.

There is also a latent correctness problem in the same area: the export worker reads `record.session.user?.username`, but `queryRecordsRaw`'s `$lookup` joins only `SessionModel` and never `UserModel`, so `$session.user.username` is always undefined and every exported row gets `'N/A'`. The source carries a `// TBD test this works` comment at `instrument-records.service.ts:510`.

### Measurements

With 50,000 sessions across 8 groups (6,250 sessions in the group being viewed), 5,000 subjects:

```
sessions {groupId}, no indexes    50,000 docs examined -> 6,250 returned, 11 ms server-side
sessions {groupId}, with index     6,250 docs examined -> 6,250 returned,  6 ms server-side
```

Server time is modest, but that is not the cost that matters here. The 6,250 sessions each embed a full subject document and are serialised to JSON, sent over the network, and then parsed by `$SessionWithUser.array().parseAsync(...)` in the browser — a full Zod validation pass over 6,250 objects — before any of it is used. The join is then O(40 × 6,250) = 250,000 comparisons to populate a 40-row table. All of this scales linearly with the group's total session count, while the data actually rendered does not grow at all.

Client, Server

### Proposed Solution

**1. Return the username with the record, not as a separate collection to join client-side.** The cleanest fix: have `InstrumentRecordsService.find()` include the session's user, so the client needs no second request and no join at all:

```ts
// instrument-records.service.ts:231
const records = await this.instrumentRecordModel.findMany({
  include: {
    instrument: false,
    session: { select: { user: { select: { username: true } } } }
  },
  where: { /* unchanged */ }
});
```

This removes the `/v1/sessions` request from the datahub entirely.

**2. If the separate endpoint is kept, scope it.** `findAllIncludeUsernames` should accept `subjectId` (and ideally `id: { in: [...] }`) and drop `include: { subject: true }`, which is never read by this caller. The client already knows exactly which `sessionId`s it needs — it has them on the records.

**3. Replace the nested `find` with a Map** regardless of which of the above lands, so the join is O(n + m):

```ts
const sessionsById = new Map(sessions.map((s) => [s.id, s]));
const records = recordsQuery.data.map((record) => {
  const username = sessionsById.get(record.sessionId)?.user?.username ?? 'N/A';
  // ...
});
```

**4. Separately, fix the export path's username.** Either add a `$lookup` on `UserModel` in `queryRecordsRaw`, or denormalise the username onto the record. Worth tracking as its own bug, but it is the same underlying "username lives one join away" problem.

### Estimated Difficulty

Medium

### Priority

High

gdevenyi:
Refs #1410. Deliberately not `Fixes`: the unbounded `GET /v1/sessions` and the export path's `N/A` username are left open.

Viewing one subject's records fetched **every session in the current group**, each with its subject document embedded, and joined them in the browser with `sessions.find(...)` inside `records.map(...)` — O(records × sessions) to label a few dozen rows with a username.

The username is now resolved on the server, keyed on the sessions the returned records actually reference.

## Measured effect

Against a real MongoDB 8 replica set, database provisioned through the app's own `POST /v1/setup`, with 20,031 sessions in the group being viewed:

| | `main` | this branch |
|---|---|---|
| `GET /v1/instrument-records?subjectId=…` | 2,624 B | 2,786 B |
| `GET /v1/sessions?groupId=…` | **12,922,313 B / 950 ms** | *not requested* |
| **total for one subject view** | **~12.9 MB** | **2.8 KB** |

~4,600× fewer bytes. The `/v1/sessions` request is gone from the datahub entirely, and the client-side join with it — `useInstrumentVisualization` now reads `record.session?.user?.username` directly.

## An important correction to the approach the issue proposed

#1410 proposed doing this with an `include` on the relation:

```ts
include: {
  instrument: false,
  session: { select: { user: { select: { username: true } } } }
}
```

**That does not work, and I only found out by testing it against a live instance.** `InstrumentRecord.session` is declared *required* in the prisma schema (`session Session @relation(...)`), so if any single record's session has since been deleted, Prisma aborts the whole query:

```
Invalid `prisma.instrumentRecord.findMany()` invocation:
Inconsistent query result: Field session is required to return data, got `null` instead.
```

I reproduced this by deleting one session and re-querying: **`main` returns 200, the `include` version returned 500.** That would have turned a performance fix into an outage for any instance with an orphaned record — and MongoDB enforces no referential integrity, so nothing prevents that state.

The merged approach is a second scoped query instead:

```ts
const sessions = await this.sessionModel.findMany({
  select: { id: true, user: { select: { username: true } } },
  where: { id: { in: Array.from(new Set(records.map((record) => record.sessionId))) } }
});
```

joined with a Map. It is bounded by the records being returned (not by the group's session count), it degrades to a missing username rather than a failed request, and it costs one extra round trip that is skipped entirely when there are no records.

## Verified on a live instance

All three cases, against the running API:

```
record 6a5e67b4 -> username: "perfadmin"   (session with a user)
record 6a5e67b4 -> username: "perfadmin"
record 6a5e67b4 -> username: null          (session with no user)
record 6a5e67b4 -> username: null          (session deleted — no crash)
HTTP 200
```

## Tests

**5 new server tests** covering the username labelling, deduplication of session ids, skipping the lookup when there are no records, and — the one that matters most — that a record whose session no longer exists is still returned, with no username.

**The 8 existing `useInstrumentVisualization` tests pass unchanged.** They assert that `Username,…testusername` appears in the CSV, TSV, Excel and JSON exports; those assertions are untouched, so they demonstrate the username still reaches every export format, just sourced from the record instead of a separate fetch. Only the fixture moved.

## Checks

- `tsc --noEmit` on `packages/schemas`, `apps/api`, `apps/web` — clean
- `eslint src` on all three — clean
- `prettier --check` — clean
- `vitest run` (whole workspace) — 257 passed, 1 skipped, 1 failed

The one failure is `instrument-records.service.spec.ts > upload > should create records with pending set`, which **fails identically on pristine `main`** and is unrelated.

## Notes for review

- `useFindSessionQuery` had exactly one caller and is removed, since this change orphaned it.
- **`GET /v1/sessions` is left in place.** It is now unused by the web client but is still unbounded (no pagination, no `subjectId` filter, and it embeds the full subject per session). Removing it is a breaking API change I did not want to make unilaterally; scoping or removing it is worth a follow-up, and #1410 remains open for that.
- I could not do a manual browser pass — Chrome in this environment is snap-confined and cannot reach host ports. The API behaviour was verified directly against the live stack as shown above.

---

Rebased onto `26ec89168`. Unit and e2e suites green locally; `tsc` and `eslint` clean.

**1. End-to-end coverage for the api → web contract.** Both existing tiers mocked the other side, so
nothing exercised the username actually travelling. `instrument-completion.spec.ts` already
administers an instrument and lands on the subject's record table, so it now asserts the Username
cell shows the user who conducted the session, via a new `subject-table-cell-<column>` testid.
Verified by mutation: returning the records unlabelled from `find` fails the spec.

`roleToken` becomes `roleAccount`, returning the seeded username alongside the token, so the spec
knows who it is acting as. It had one caller (`authenticateAs`). `testing/AGENTS.md` updated to match.

**2. `Fixes #1410` → `Refs #1410`**, since the unbounded `GET /v1/sessions` and the export path's
`N/A` username are deliberately left open.

**3. The username lookup no longer runs on every `find`.** `/dashboard` fetches every record in the
group and reads `instrumentId` alone, so it was paying for a second query with an `$in` as long as
the group's record list to produce a field it discards. Gated on `subjectId`, which only the
per-subject datahub view supplies.

**4. The `accessibleQuery` dependency is pinned.** Confirmed your finding independently — given an
ability with no `read Session` rule at all, it throws a CASL `ForbiddenError` rather than returning a
restrictive filter, so the request would 500 rather than 403. Nothing can reach that today, and
nothing pinned it either. A test in `ability.factory.test.ts` now asserts every base permission level
grants `read Session`, so removing one fails there rather than in production, and the call site
records what it depends on.

Copilot's three comments all asked for the same ability scoping, which the second commit already
applied.


joshunrau:
@gdevenyi please rebase against main

joshunrau (changes requested):
Needs to apply ability scoping. Right now, this does not apply access controls.

joshunrau (changes requested):
Nice piece of work — the measurement, and especially the finding that the `include` the issue proposed 500s on an orphaned record, are the valuable parts of this. The scoping fix in the second commit is right, and `Session.user` being optional means the nested `user` select can't hit the same required-relation crash. Four things before this goes in:

1. **This needs an end-to-end test.** Both existing test layers mock the other side — the api tests mock Prisma, the `useInstrumentVisualization` test mocks `useInstrumentRecords` — so the actual api→web contract carrying the username isn't covered anywhere. `testing/src/specs/instrument-completion.spec.ts` already lands on `/datahub/**/table` and picks the instrument at line 44; asserting the Username cell shows the logged-in user there should be a few lines plus a testid.

2. **The PR body says `Fixes #1410`**, which will close the issue on merge — but you're deliberately leaving the unbounded `GET /v1/sessions` and the export path's `'N/A'` username (`instrument-records.service.ts:539`) open. Please change it to `Refs #1410`; your commit message already has it right.

3. **The username join runs on every call to `find()`**, including `/dashboard`. `apps/web/src/routes/_app/dashboard.tsx:35` fetches every record in the group and reads only `instrumentId`, so that page now pays for a second query with an `$in` as long as the group's record list, for a field it throws away. `withSessionUsernames` (`instrument-records.service.ts:278`) is unconditional — consider gating the lookup on `subjectId` being present, or on an explicit query param.

4. **Worth a test, and possibly a comment:** `accessibleQuery(ability, 'read', 'Session')` (`instrument-records.service.ts:598`) throws a CASL `ForbiddenError` rather than returning an empty filter when the ability has no `read Session` rule at all, and that surfaces as a 500 rather than a 403. I verified this directly against `@casl/prisma@1.6.2`. Nothing can trigger it today since every base permission level grants `read Session`, but `GET /v1/instrument-records` now quietly depends on that and nothing pins it down.

joshunrau:
**Suggested model: Opus 5**

Sizing the remaining work on this PR for whoever picks it up, human or agent. The action items span `apps/api` and `testing/`, and item 3 is a design choice — gate the session lookup on `subjectId` or on an explicit query param — rather than a specified instruction. That puts it above a mechanical tier.

joshunrau:
This is in good shape — all four items from the last round are properly addressed, not just
nominally. I re-ran the new spec against a live stack and then mutated `find` to return the records
unlabelled: it fails on the `N/A` fallback, so the api→web contract really is pinned now, which was
the gap. The `subjectId` gate and the `accessibleQuery` pin both read well, and the comment at the
call site saying what it depends on is the right call.

Two small things, neither blocking:

1. `$InstrumentRecord`'s new `session.user` shape is character-for-character `$SessionWithUser`'s
   `user` (`packages/schemas/src/session/session.ts:29`). `packages/schemas/AGENTS.md` asks for
   shared shapes to be composed rather than re-declared, since re-declaring is how the two drift —
   `session: z.object({ user: $SessionWithUser.shape.user }).nullish()` would do it.

2. The new assertion hardcodes `roleAccount('GROUP_MANAGER')` while the page itself is driven as
   `actingRole`. They line up only because the default is `GROUP_MANAGER`; if this file ever picks up
   a `test.use({ actingRole: ... })` the assertion would quietly compare against the wrong account.
   Taking `actingRole` from the fixtures keeps the two in step.

One thing worth knowing rather than fixing: your `ability.factory.test.ts` pin covers the three base
permission levels, but `basePermissionLevel` is nullable, so a user with no base level and only
`additionalPermissions: [{ action: 'read', subject: 'InstrumentRecord' }]` — reachable through
`PATCH /v1/users/:id` though not through the web UI — clears `@RouteAccess` and then hits the
`ForbiddenError` in `withSessionUsernames`, i.e. a 500. Not something this PR introduces the shape
of, and a loud failure rather than a quiet one, so I would not hold the PR for it.

If you hand this to Claude Code, **Sonnet 5** is the right size — two mechanical edits in files the
items name outright.

Reviewed at commit 5ae02a8002a248d2f2983a5989c6d948bdf39cb9.

Issue #1406, opened 2026-07-20
Performance: no indexes are ever created on any MongoDB collection


gdevenyi:
### Current Implementation and Limitations

`apps/api/prisma/schema.prisma` declares **no indexes at all** — there is not a single `@@index` in the file, and the only `@unique` attributes are `Group.name`, `InstrumentRepo.url` and `InstrumentRecord.assignmentId`.

Worse, those unique indexes are never created either, because **the API never runs `prisma db push`**:

- `apps/api/package.json` has scripts `build`, `db:generate`, `dev`, `dev:test`, `format`, `lint`, `start`, `test` — there is no `db:push`.
- `turbo.json` attaches `db:push` to `gateway#build` only; the generic `build` task depends on `["^build", "db:generate"]`.
- `apps/api/Dockerfile` runs `turbo build --filter=@opendatacapture/api`, so nothing ever pushes the schema.

MongoDB auto-creates a collection on first insert with only the default `_id_` index. The result is that **every query that filters on anything other than `_id` is a full collection scan** (`COLLSCAN`), including the queries on the hottest paths:

| Call site | Filter fields |
|---|---|
| `InstrumentRecordsService.find` (`instrument-records.service.ts:231`) | `subjectId`, `instrumentId`, `groupId`, `date`, `pending` |
| `InstrumentRecordsService.count` (`:69`) | `groupId`, `date` |
| `SessionsService.findAllIncludeUsernames` (`sessions.service.ts:100`) | `groupId` |
| `SubjectsService.querySubjectIdsWithRecords` (`subjects.service.ts:162`) | `groupId` |
| `AuditService.find` (`audit.service.ts:11`) | `groupId`, `userId`, `action`, `entity` |
| `FilesService`, `AssignmentsService.find`, etc. | `recordId`, `subjectId` |

### Measurements

MongoDB 8, collections built to mirror this schema exactly with no indexes (as deployed): 5,000 subjects, 50,000 sessions, 200,000 instrument records (260 MiB), 120,000 audit logs. All measurements are on a warm cache with the whole working set in RAM, i.e. this is the **best case** — once the collection exceeds available RAM a `COLLSCAN` degrades to disk I/O and the gap widens by orders of magnitude.

**As deployed (no indexes):**

```
query                                                docsExamined  returned    time  stage
records {subjectId,instrumentId,groupId,date,pending}     200,000        40    72ms  COLLSCAN
[...]

PR #1423, opened 2026-07-20
fix(api): resolve the instrument subject filter without a relation join

### LINKED ISSUE #1416: Datahub subject view returns HTTP 500 once one instrument exceeds ~182k records (Prisma relation filter hits MongoDB's 100 MiB $lookup ceiling)

`GET /v1/instruments/info?subjectId=<id>` **returns HTTP 500 and the datahub subject view stops working entirely** once a single instrument accumulates roughly 182,000 records. This is a hard, permanent failure, not a slowdown, and it is reached silently as an instance grows.

The cause is the relation filter in `InstrumentsService.find()` (`apps/api/src/instruments/instruments.service.ts:118`):

```ts
const instruments = await this.instrumentModel.findMany({
  where: {
    AND: [
      {
        records: query.subjectId ? { some: { subjectId: query.subjectId } } : undefined
      },
      accessibleQuery(ability, 'read', 'Instrument')
    ]
  }
});
```

On MongoDB, Prisma compiles `records: { some: {...} }` into a `$lookup` from `InstrumentModel` into `InstrumentRecordModel` that materialises **every record belonging to that instrument into a single array field**, and only then applies the `subjectId` predicate. MongoDB enforces a hard 100 MiB (104,857,600 byte) ceiling on the size of a single document's `$lookup` output array. Once the busiest instrument's records exceed that, the aggregation aborts:

```
Kind: Command failed: Error code 4568 (Location4568):
Executor error during aggregate command on namespace:
<db>.InstrumentModel :: caused by :: Total size of documents in
InstrumentRecordModel matching pipeline's $lookup stage exceeds 104857600 bytes
```

This limit cannot be raised — `allowDiskUse` does not apply to it.

The blast radius is the core clinical workflow. `useInstrumentVisualization` (`apps/web/src/hooks/useInstrumentVisualization.ts:47`) calls this endpoint on mount:

```ts
const instrumentInfoQuery = useInstrumentInfoQuery({
  params: { allEditions: true, kind: params.kind, subjectId: params.subjectId }
});
```

and that hook backs **both** `/datahub/$subjectId/table` and `/datahub/$subjectId/graph`. When it 500s, neither page can render — there is no partial degradation and no workaround available to the user.

Critically, the threshold is per-instrument, not per-collection, so it is reached by the *most-used* instrument first. In a real clinical deployment records are not evenly distributed: a core intake or demographics instrument administered at every visit accumulates records far faster than the rest. 182,000 records on one instrument is 5,000 subjects × 36 visits, or 15,000 subjects × 12 visits — an ordinary size for this application.

### Measurements

Measured against a live stack: the API running from `apps/api`, a real MongoDB 8 replica set, and a database provisioned through the application's own `POST /v1/setup` endpoint, then grown to 5,000 subjects / 50,400 sessions / 204,800 instrument records (112 MiB). Records distributed so that one "core intake" instrument holds 40% and the remaining 30 instruments share the rest.

**Endpoint behaviour as the busiest instrument grows:**

| Records on busiest instrument | Size | `GET /v1/instruments/info?subjectId=` |
|---|---|---|
| 81,920 | 45 MiB (45% of ceiling) | **3,592 ms** for a 22 KiB response |
| 81,920 + `{instrumentId: 1}` index | 45 MiB | 495 ms |
| 204,800 | 112 MiB (over ceiling) | **HTTP 500** |

So the endpoint is already pathological well below the limit — 3.6 seconds to return 22 KiB — and then fails outright.

Reproduced directly to confirm the mechanism is the `$lookup` array and not something else:

```js
db.InstrumentModel.aggregate([
  { $lookup: { from: 'InstrumentRecordModel', localField: '_id',
               foreignField: 'instrumentId', as: 'records' } },
  { $match: { 'records.subjectId': <someSubjectId> } },
  { $project: { _id: 1 } }
]);
// MongoServerError: Used too much memory for a single array.
// Memory limit: 104857600 bytes. The array contains 179503 elements
```

The same request without `subjectId` (`GET /v1/instruments/info`) returns in 29 ms, confirming the relation filter is the sole cause.

Server

### Proposed Solution

**1. Do not use a relation filter here.** The query only needs "which instruments does this subject have records for", which is a one-field distinct over `InstrumentRecord` — a query against the child collection rather than a join from the parent:

```ts
async find<TKind extends InstrumentKind>(query: InstrumentQuery<TKind> = {}, { ability }: EntityOperationOptions = {}) {
  let instrumentIdFilter: Prisma.InstrumentWhereInput | undefined;
  if (query.subjectId) {
    const ids = await this.instrumentRecordModel
      .findMany({ distinct: ['instrumentId'], select: { instrumentId: true }, where: { subjectId: query.subjectId } })
      .then((records) => records.map((record) => record.instrumentId));
    instrumentIdFilter = { id: { in: ids } };
  }
  const instruments = await this.instrumentModel.findMany({
    where: { AND: [instrumentIdFilter ?? {}, accessibleQuery(ability, 'read', 'Instrument')] }
  });
  // ...
}
```

With the `{subjectId, instrumentId, date}` index proposed in the indexing issue this touches only that subject's records (~40 documents) instead of the whole instrument's record set. It has no size ceiling and is O(records for this subject) rather than O(records for this instrument).

**2. Add `@@index([instrumentId])` on `InstrumentRecord`** regardless. It takes the current implementation from 3,592 ms to 495 ms, which buys time even before the query is restructured — though it does **not** remove the 100 MiB ceiling, so it is a mitigation and not a fix.

**3. Add a regression test at scale.** A test that seeds more records than the ceiling for one instrument and asserts this endpoint still responds would have caught this. Because the failure only appears above a data-volume threshold, no amount of small-fixture testing will surface it.

**4. Audit for other relation filters on MongoDB.** `records: { some: ... }` is the instance found here; any other `some`/`every`/`none` filter across a high-cardinality relation has the same ceiling. `FilesService.find` (`instrument-records/files/files.service.ts:38`) uses `files: { every: ... }`, which is bounded by files-per-record and so is safe today, but the pattern is worth a sweep.

### Estimated Difficulty

Medium

### Priority

High

gdevenyi:
Fixes #1416.

`GET /v1/instruments/info?subjectId=` returns **HTTP 500** — and the datahub subject view stops rendering entirely — once a single instrument accumulates roughly 182,000 records.

## Cause

`InstrumentsService.find` expressed "instruments this subject has records for" as a prisma relation filter:

```ts
records: query.subjectId ? { some: { subjectId: query.subjectId } } : undefined
```

On MongoDB, prisma compiles that into a `$lookup` which materialises **every record belonging to an instrument** into a single array before applying the predicate. MongoDB caps one document's `$lookup` output at 100 MiB and aborts the whole aggregation past it (error 4568):

```
Executor error during aggregate command on namespace: <db>.InstrumentModel
:: caused by :: Total size of documents in InstrumentRecordModel matching
pipeline's $lookup stage exceeds 104857600 bytes
```

The limit cannot be raised, and `allowDiskUse` does not apply to it.

Two things make this worse than the raw number suggests:

- **The threshold is per-instrument, not per-collection**, so the *busiest* instrument hits it first. A core intake questionnaire administered at every visit reaches 182,000 records at 5,000 subjects × 36 visits — an ordinary size here.
- **There is no partial degradation.** `useInstrumentVisualization` calls this endpoint on mount, and it backs both `/datahub/$subjectId/table` and `/datahub/$subjectId/graph`, so both pages fail and the user has no workaround.

## The fix

Query the child collection instead of joining from the parent. The work is then bounded by the *subject's* records (tens) rather than the *instrument's* (hundreds of thousands), and there is no size ceiling:

```ts
const subjectInstrumentIds = query.subjectId ? await this.findInstrumentIdsBySubject(query.subjectId) : null;

const instruments = await this.instrumentModel.findMany({
  where: {
    AND: [subjectInstrumentIds ? { id: { in: subjectInstrumentIds } } : {}, accessibleQuery(ability, 'read', 'Instrument')]
  }
});
```

## Verified against a live instance

Real API, real MongoDB replica set, database provisioned through the app's own `POST /v1/setup`, then one instrument pushed past the ceiling.

**Above the ceiling — busiest instrument at 105.7 MiB, identical database for both:**

| | result |
|---|---|
| `main` | **HTTP 500** (three times), `exceeds 104857600 bytes` |
| this branch | **HTTP 200** in 77 ms, 811 bytes |

**Below the ceiling — same 55.3 MiB instrument, where `main` still answers:**

| | latency |
|---|---|
| `main` | 1.83s / 1.69s / 1.66s |
| this branch | 0.10s / 0.042s / 0.041s |

So this is not only an outage fix: even well under the limit the endpoint was taking **~1.7 seconds to return 811 bytes**, and it now takes ~40 ms.

## Tests

Three added to `instruments.service.spec.ts`:

- the subject filter is resolved against `InstrumentRecord` and the instrument query contains no `records` relation filter (this is the regression guard — it asserts the *absence* of the construct that fails);
- no record query is issued when no subject is given;
- a subject with no records yields `{ id: { in: [] } }` rather than an unfiltered query.

The existing `findInfo` tests are unchanged and still pass.

## Checks

- `eslint src` — clean
- `prettier --check` — clean
- `vitest run` (whole workspace) — 255 passed, 1 skipped, 1 failed

The one failure is `instrument-records.service.spec.ts > upload > should create records with pending set`. This branch is cut from `main`, so it still carries that pre-existing failure — **#1422 fixes it**, and the suite is fully green there.

## Notes

- `@@index([instrumentId])` on `InstrumentRecord` (proposed in #1406) is a useful mitigation for the old query but **not a fix** — it does not remove the 100 MiB ceiling. This change removes the ceiling from the picture entirely; the index still helps the new query.
- Because the failure only appears above a data-volume threshold, no small-fixture test would have caught it. A seeded at-scale test in the e2e suite would be worth adding separately.
- I swept for other relation filters on high-cardinality relations. `FilesService.find` uses `files: { every: ... }`, which is bounded by files-per-record and is safe today; it is the only other instance.

---

Rebased onto `26ec89168`. The stale CI is gone and the branch is mergeable; `lint-and-test` is green.

**Conflict resolution.** `main` gained a `seriesGroupId` clause in the same `where` this PR rewrites.
Both are kept — the `records: { some: ... }` relation filter is the only thing removed, replaced by
the resolved id list. The spec conflicts were pure additions from `main` (series fixtures), kept as-is.

**Copilot: the record lookup was unscoped.** Fixed. `findInstrumentIdsBySubject` now takes the
caller's ability and applies `accessibleQuery(ability, 'read', 'InstrumentRecord')`. This mattered:
the ids feed straight into the instrument query, so an unscoped lookup would disclose which
instruments a subject had been administered outside the caller's groups — something the relation
filter it replaces did not do. Pinned by a test asserting the exact `accessibleQuery` clause, using a
*conditioned* ability so it cannot pass vacuously.

**Copilot: misleading test name.** Renamed to say what it actually asserts — that no *record* query
runs when no subject is given.

---

## Round 2: rebased onto `5bfe4b2dc`, review items addressed

**1. The new 500 — confirmed and fixed.** You were right, and I reproduced it exactly: with an
ability holding no `read InstrumentRecord` rule, `accessibleQuery` throws
`ForbiddenError: It's not allowed to run "read" on "InstrumentRecord"` rather than returning `{}`.
Only an `undefined` ability returns `{}`. `findInstrumentIdsBySubject` now returns an empty id list
for such a caller, before `accessibleQuery` is reached. The scoping itself stays, as you asked.

**2. A test that would have caught it.** Added, built through
`AbilityFactory.createForPayload({ basePermissionLevel: 'STANDARD', ... })` rather than by hand —
your point about the existing test is exactly right, a hand-built ability tends to carry the very
rule whose absence breaks this. Verified by mutation: removing the guard fails it with that
`ForbiddenError`.

**3. End-to-end.** Added to `authorization.spec.ts`: a STANDARD case asserting the endpoint answers
200 rather than 500, and a group-manager case alongside it. The STANDARD one returns 500 without the
guard — verified. The at-scale reproduction is noted in the spec as a deliberate gap rather than left
silent, as you suggested; the populated UI path is already covered by `instrument-completion.spec.ts`,
so I pointed at that rather than duplicating it.

Noted on `@@index([subjectId])` — agreed it is the natural companion, and I have left it out of this
PR as you asked. The group-visibility consequence stays as decided.

### DIFF
--- a/.agents/docs/architecture/auth-and-permissions.md
+++ b/.agents/docs/architecture/auth-and-permissions.md
@@ -70,6 +70,16 @@ available in this codebase. Check the `where` clause of every query you add or t
 that the controller actually forwards `@CurrentUser('ability')` — `EntityOperationOptions.ability`
 is optional, so a controller that simply never passes it compiles and runs.
 
+**A defined ability holding no rule for the queried subject does not produce a deny-all filter —
+CASL's `accessibleBy` throws `ForbiddenError`.** That error is not an `HttpException`, so the
+global exception filter renders it as a 500. It is reachable whenever the route guard names a
+different subject than the service queries: a STANDARD user passes a `read Instrument` guard but
+holds no rule on `InstrumentRecord`, so a record query made on their behalf throws. The throw is
+deliberate — `accessibleQuery` is kept as-is rather than returning a deny-all filter — so where
+such a caller is legitimate, guard the query with `ability.can(action, subject)` and return the
+empty result instead; `findInstrumentIdsBySubject` in `src/instruments/instruments.service.ts` is
+the reference example.
+
 One call site uses `{ ...accessibleQuery(...), id: subjectId }` instead of `AND: [...]`
 (`src/subjects/subjects.service.ts`). Spreading merges keys and will silently lose a condition if
 the generated query and the literal share one. Prefer `AND`.
--- a/apps/api/AGENTS.md
+++ b/apps/api/AGENTS.md
@@ -83,6 +83,14 @@ restriction**, which reads clinical data across every group. Nothing catches thi
 eslint, not a passing test suite. It is the single highest-severity mistake available in this
 codebase, so verify the `where` clause of any query you add or edit.
 
+The defined case has the opposite failure mode: **an ability holding no rule at all for the queried
+subject makes `accessibleQuery` throw CASL's `ForbiddenError`**, which is not an `HttpException`
+and so surfaces as a 500. This is reachable whenever a route's guard names one subject but the
+service queries another — a STANDARD user passes `read Instrument` yet holds no `read` rule on
+`InstrumentRecord`. That behaviour is deliberate and stays; where such a caller is legitimate,
+short-circuit before the query with `ability.can(action, subject)` and return the empty result —
+see `findInstrumentIdsBySubject` in `src/instruments/instruments.service.ts`.
+
 Granting a new `action`/`subject` pair means editing `src/auth/ability.factory.ts` and adding tests
 for **both** the allow and the deny case — see `src/auth/__tests__/ability.factory.test.ts`.
 
--- a/apps/api/src/instruments/__tests__/instruments.service.spec.ts
+++ b/apps/api/src/instruments/__tests__/instruments.service.spec.ts
@@ -9,6 +9,7 @@ import type { SeriesInstrument } from '@opendatacapture/runtime-core';
 import type { WithID } from '@opendatacapture/schemas/core';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { AbilityFactory } from '@/auth/ability.factory';
 import { accessibleQuery, createAppAbility } from '@/auth/ability.utils';
 
 import { InstrumentsService } from '../instruments.service';
@@ -690,6 +691,25 @@ describe('InstrumentsService', () => {
 
       const [call] = instrumentRecordModel.findMany.mock.lastCall as [{ where: { AND: unknown[] } }];
       expect(call.where.AND[0]).toStrictEqual(accessibleQuery(ability, 'read', 'InstrumentRecord'));
+    });
+
+    // Built through the factory rather than by hand: a hand-built ability tends to include a
+    // `read InstrumentRecord` rule, and it is the absence of one that breaks. A STANDARD user holds
+    // `create` but not `read`, and this route is gated on `read Instrument`, which they do hold.
+    it('should resolve for a caller who may read no records, rather than failing the request', async () => {
+      const abilityFactory = new AbilityFactory(MockFactory.createMock(LoggingService) as unknown as LoggingService);
+      const ability = abilityFactory.createForPayload({
+        basePermissionLevel: 'STANDARD',
+        groups: [{ id: 'group-1' }],
+        id: 'user-1'
+      } as any);
+
+      await expect(instrumentsService.find({ subjectId: 'subject-1' }, { ability })).resolves.toStrictEqual([]);
+
+      expect(instrumentRecordModel.findMany).not.toHaveBeenCalled();
+      expect(instrumentModel.findMany.mock.lastCall?.[0]).toMatchObject({
+        where: { AND: expect.arrayContaining([{ id: { in: [] } }]) }
+      });
     });
 
     it('should return nothing when the subject has no records', async () => {
--- a/apps/api/src/instruments/__tests__/instruments.service.spec.ts
+++ b/apps/api/src/instruments/__tests__/instruments.service.spec.ts
@@ -9,6 +9,8 @@ import type { SeriesInstrument } from '@opendatacapture/runtime-core';
 import type { WithID } from '@opendatacapture/schemas/core';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { accessibleQuery, createAppAbility } from '@/auth/ability.utils';
+
 import { InstrumentsService } from '../instruments.service';
 
 import type { InstrumentVirtualizationContext } from '../instruments.service';
@@ -642,6 +644,62 @@ describe('InstrumentsService', () => {
         { id: 'owned', seriesGroupId: 'group-1' },
         { id: 'shared', seriesGroupId: null }
       ]);
+    });
+  });
+
+  describe('find', () => {
+    beforeEach(() => {
+      instrumentModel.findMany.mockResolvedValue([]);
+      instrumentRecordModel.findMany.mockResolvedValue([]);
+      vi.spyOn(instrumentsService as any, 'instantiate').mockResolvedValue([]);
+    });
+
+    it('should not query records when no subject is given, so the unfiltered listing costs one query', async () => {
+      await instrumentsService.find();
+      expect(instrumentRecordModel.findMany).not.toHaveBeenCalled();
+    });
+
+    // A `records: { some: ... }` relation filter compiles to a $lookup that materialises every
+    // record belonging to an instrument, which mongodb aborts past 100 MiB. The subject's own
+    // records are queried instead, so the work is bounded by the subject rather than the instrument.
+    it('should resolve the subject filter against records rather than joining from instruments', async () => {
+      instrumentRecordModel.findMany.mockResolvedValueOnce([{ instrumentId: 'id-1' }, { instrumentId: 'id-2' }] as any);
+
+      await instrumentsService.find({ subjectId: 'subject-1' });
+
+      expect(instrumentRecordModel.findMany).toHaveBeenCalledWith({
+        distinct: ['instrumentId'],
+        select: { instrumentId: true },
+        where: { AND: [{}, { subjectId: 'subject-1' }] }
+      });
+      expect(instrumentModel.findMany.mock.lastCall?.[0]).toMatchObject({
+        where: { AND: expect.arrayContaining([{ id: { in: ['id-1', 'id-2'] } }]) }
+      });
+      expect(JSON.stringify(instrumentModel.findMany.mock.lastCall?.[0])).not.toContain('records');
+    });
+
+    it('should constrain the record lookup to what the caller may read, so the filter cannot be resolved from other groups records', async () => {
+      // Conditions are what make this meaningful: an unconditional read rule yields `{}`, which is
+      // indistinguishable from the ability never having been applied.
+      const ability = createAppAbility([
+        { action: 'read', conditions: { groupId: { in: ['group-1'] } }, subject: 'InstrumentRecord' },
+        { action: 'read', subject: 'Instrument' }
+      ]);
+
+      await instrumentsService.find({ subjectId: 'subject-1' }, { ability });
+
+      const [call] = instrumentRecordModel.findMany.mock.lastCall as [{ where: { AND: unknown[] } }];
+      expect(call.where.AND[0]).toStrictEqual(accessibleQuery(ability, 'read', 'InstrumentRecord'));
+    });
+
+    it('should return nothing when the subject has no records', async () => {
+      instrumentRecordModel.findMany.mockResolvedValueOnce([]);
+
+      await instrumentsService.find({ subjectId: 'subject-with-no-records' });
+
+      expect(instrumentModel.findMany.mock.lastCall?.[0]).toMatchObject({
+        where: { AND: expect.arrayContaining([{ id: { in: [] } }]) }
+      });
     });
   });
 });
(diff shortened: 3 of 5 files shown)


joshunrau:

**Failing check:** `lint-and-test` — [run 29770064355](https://github.com/DouglasNeuroInformatics/OpenDataCapture/actions/runs/29770064355/job/88445677061)

```
FAIL  api  src/instrument-records/__tests__/instrument-records.service.spec.ts
  > InstrumentRecordsService > upload > should create records with pending set,
    so they are not excluded from queries filtering on the field
AssertionError: expected "vi.fn()" to be called with arguments: [ { data: [ ObjectContaining{ pending: false } ] } ]
```

That test is not in your diff and no longer exists on `main` — it was removed in `722f0a0b0` ("test(api): update stale pending-on-create assertion for instrument records", 2026-07-20 22:39 EDT), which landed after this run was created. The run predates the fix, so the result is stale.

**Also:** GitHub reports this branch as `CONFLICTING` with `main`.

Please rebase or merge `main` into `fix/instruments-subject-filter-lookup`, resolve the conflicts, and push. That will both clear the conflict and re-run CI against the corrected `main`. I'll pick the review back up once it's green.

joshunrau:
**Suggested model: Opus 5**

Sizing the remaining work on this PR for whoever picks it up, human or agent. Beyond the stale CI, the conflicts land in a service `main` has since changed, so resolution needs the query semantics held in mind rather than a textual merge.

joshunrau (changes requested):

**Blocking:** conflicts with `main`, plus stale CI. Details in [my earlier comment](https://github.com/DouglasNeuroInformatics/OpenDataCapture/pull/1423#issuecomment-5094051128) — the failing test isn't in your diff and no longer exists on `main`. Rebase `fix/instruments-subject-filter-lookup`, resolve the conflicts, and push; that clears both.

No content review has been done on this branch yet — it's gated on CI being green.

joshunrau (changes requested):
Thanks for this — the diagnosis is correct and the `$lookup` ceiling is a real outage, so the
direction is right and the measurements are appreciated. Three things before it can land.

1. The scoped record lookup introduces a new 500 on the endpoint you are fixing.
   `accessibleQuery(ability, 'read', 'InstrumentRecord')` in
   `apps/api/src/instruments/instruments.service.ts:537` does not return an empty filter when the
   caller holds no `read InstrumentRecord` rule — it throws
   `ForbiddenError: It's not allowed to run "read" on "InstrumentRecord"`. Only an `undefined` ability
   returns `{}`. `basePermissionLevel: 'STANDARD'` grants `create` but not `read` on `InstrumentRecord`
   (`apps/api/src/auth/ability.factory.ts:47-52`), and `GET /v1/instruments/info` is gated on
   `read Instrument`, which STANDARD does hold. libnest's `GlobalExceptionFilter` turns a
   non-`HttpException` into a 500, so a STANDARD user calling that endpoint with a `subjectId` now gets
   an Internal Server Error where they previously got 200. Every other
   `accessibleQuery(..., 'InstrumentRecord')` call site in the repo sits behind a matching
   `@RouteAccess({ action: 'read', subject: 'InstrumentRecord' })`; this one does not. Please return an
   empty id list for a caller who cannot read records rather than letting the error escape.

2. Add a test that would have caught it, in
   `apps/api/src/instruments/__tests__/instruments.service.spec.ts` alongside the scoping test at line
   681. Build the ability with `AbilityFactory.createForPayload({ basePermissionLevel: 'STANDARD', ... })`
   and assert `find({ subjectId })` resolves. The current scoping test hand-builds an ability that
   happens to include a `read InstrumentRecord` rule, so it passes either way.

3. An end-to-end test in `testing/` is required for every change here. A spec loading
   `/datahub/$subjectId/table` as a GROUP_MANAGER and asserting the instrument list renders, plus a
   STANDARD case in `src/specs/authorization.spec.ts` asserting the endpoint does not return a 500,
   would cover both the fix and point 1. You are right that the at-scale reproduction cannot be
   expressed there — note that in the spec so the gap is deliberate rather than missing.

One more thing, for a separate PR rather than this one: `InstrumentRecord` has no `@@index` at all, so
the new lookup is a collection scan filtered by `subjectId`. It is still far better than the `$lookup`
it replaces and it has no ceiling, but `@@index([subjectId])` is the natural companion to this change.

On the visibility question: adding the record-level scoping means that for a subject enrolled in more
than one group, instruments administered in another group disappear from a group manager's datahub
list. Joshua has confirmed he wants that — it is a disclosure fix and it matches
`/v1/instrument-records` — so keep the scoping, and fix point 1 by returning an empty id list rather
than by removing it.

If you hand this to Claude Code, **Fable 5** is the right size — the fix turns on CASL ability
semantics, route-guard alignment and group-scoping intent at once, on the auth and data-integrity path.

Reviewed at commit 3c3e25f9d74e92a834b11ddafb5b00caa260bfaa.

joshunrau (changes requested):
Thanks — this is in good shape. The empty-id-list guard, the `AbilityFactory`-built unit test and the
STANDARD end-to-end case all landed as asked, and you are right that `instrument-completion.spec.ts`
already covers the populated group-manager path: it picks the instrument out of the datahub combobox,
which is exactly this endpoint. Lint is clean and the api suite is green here. Two things.

1. `testing/src/specs/authorization.spec.ts:68-80` — the group-manager case asserts an empty list for
   `subjectId=any-subject`, a subject that does not exist, so it passes whether the scoping works or
   is broken for every caller. The title says it "should serve the subject-filtered instrument list",
   which it cannot demonstrate. Either seed a record through the `api` fixture and assert that
   instrument comes back, or rename it to what it does assert — that the endpoint answers rather than
   erroring.
2. The `accessibleQuery` behaviour your guard depends on is documented nowhere.
   `.agents/docs/architecture/auth-and-permissions.md:66-71` and `apps/api/AGENTS.md:81-84` describe
   only the `undefined` ability → `{}` case. Please add the other one to those paragraphs: a defined
   ability holding no rule for the subject throws `ForbiddenError`, which is not an `HttpException`
   and so renders as a 500. That is the whole reason `findInstrumentIdsBySubject` needs its
   short-circuit, and it should not have to be rediscovered.

On the broader question your guard raises — whether `accessibleQuery` should return a deny-all filter
rather than throwing — the decision is to leave it as it is. Every current call site was checked and
none is reachable by a caller holding a defined ability with no rule for the subject it queries, so
the per-call-site guard stays the pattern. Nothing to do for this PR beyond item 2, which is what
keeps that from being rediscovered the hard way.

If you hand this to Claude Code, **Fable 5** is the right size — the work sits on the auth and
group-scoping path, and the documentation item means writing the normative description of the repo's
most safety-critical helper.

Reviewed at commit 40e51f5d3fbca8b0b6940b6c797f72971423e2ba.

joshunrau (approved):

PR #1434, opened 2026-07-23
feat: add mailer, Spanish language support, and UX improvements

thomasbeaudry:
## Summary

This PR adds email/mailer functionality, Spanish language support, and several UX improvements to Open Data Capture.

### Mailer

- **Email remote assignments**: after creating a remote assignment, clinicians can email the link directly to a participant with a selected template and language
- **Email new users**: welcome emails can be sent to newly created users with a configurable template
- **Admin mail settings page**: admins can configure SMTP settings and test the connection
- **Group email templates page**: group managers can create, edit, and delete email templates for remote assignments and information categories
- **Default template viewer**: built-in default template can be previewed with a language dropdown toggle
- **Template validation**: duplicate template names are rejected; missing translation warnings prompt confirmation before saving
- **"Email language" tooltip**: explains that emails and assignments are sent in the selected language when available, and subjects can still choose their preferred language on the gateway
- **Email language dropdown filtered by instrument**: only shows languages the selected instrument supports
- **Set default / Default labels**: replaced "Active" / "Set active" with "Default" / "Set default" on template rows
- **SaveStatus autosave indicator** on the email templates page
- **Scroll to top** after adding a new template
- **Email Templates nav item** hidden when mail is not enabled

### Spanish Language Support

- Added Spanish (`es`) translations throughout all new and existing components using the inline `t({ en: '...', es: '...', fr: '...' })` pattern
- Added admin settings page to enable/disable individual languages via an `activeLanguages` array on setup state
- Language picker across the app respects the active languages setting
- Added Spanish support to the gateway (shows available languages for the form being completed)
- Adding future languages requires adding translations to inline `t()` calls and the `ALL_LANGUAGES` map — no schema or architecture changes needed

### UX Improvements

- **Auto-focus search bar** on the administer instrument and remote assignment pages
- **Keyboard navigation** on instrument showcase: arrow keys move highlight, Enter selects the highlighted instrument
- **Create Remote Assignment dialog** submits on Enter (replaced date picker Form component with a plain text input)
- **Admin panel auto-collapses** when navigating to non-admin pages
- **Gateway language detection** from the `?lang=` URL parameter so remote assignments open in the correct language

### Other

- Remote assignment URL now sets the form language automatically on the gateway

## Test plan

- [ ] Configure SMTP settings on the admin mail page and send a test email
- [ ] Create a remote assignment and email the link to a participant — verify email arrives in the selected language
- [ ] Create a new user with a welcome email — verify it sends
[...]

PR #1438, opened 2026-07-23
feat(i18n): add Spanish language support and per-instance active languages

thomasbeaudry:
Splits #1434 into three PRs. **This is PR 3 of 3 and merges last**, per @gdevenyi's request.

## Merge order

| Order | PR | Branch | Base |
|---|---|---|---|
| 1st | #1440 — UX fixes | `split/ux` | `main` |
| 2nd | #1439 — mailer | `split/mailer` | `main` |
| **3rd** | this PR — Spanish + active languages | `split/language` | `split/mailer` |

> **Base is `split/mailer`, not `main`.** GitHub will retarget it to `main` automatically once #1439 merges. The diff shown here is language-only.

## Why this one is stacked rather than independent

Declaring `es` in `LanguageOptions` makes `es` a **required** key in every `t({ ... })` call in `apps/web` — that's why this PR touches 82 files. Whichever PR lands *after* that declaration exists must supply Spanish for its own strings.

Since this goes last, it carries the Spanish for the mailer's strings too: **117 of the lines here translate the mail admin page, group email templates, and assignment email form** introduced in #1439. That's only possible if those files exist on this branch, hence the stack. The upside of this ordering is that #1439 needs no follow-up patch, and the `activeLanguages` / `ALL_LANGUAGES` plumbing is no longer duplicated across two branches — it lives in #1439 and is simply inherited here.

## What's here

- `es` declared in `LanguageOptions`, and Spanish added to every `t({ ... })` call in the app, plus the `es` keys in `apps/web/src/translations/*.json`. No new translation keys — the JSON changes are purely `es` additions.
- An admin **Languages** card so an instance chooses which languages it offers. The sidebar, navbar, and login language toggles render only active languages and disappear entirely when one language is active.
- The gateway resolves a `?lang` query param, which is what makes the assignment links emailed by #1439 open in the participant's language.
- Widening `Language` with `es` makes it broader than the instrument runtime's `en | fr`, so the instrument translation call sites narrow explicitly.

## Note on `admin/settings.tsx`

The autosave refactor + `SaveStatus` component landed here rather than in the UX PR, because the new Languages card is built on `autosave` — separating them would have meant inventing a Save-button variant of the card that exists nowhere in #1434.

## Verification

`tsc` and `eslint` clean for `schemas`, `react-core`, `web`, `gateway`, `api`; `pnpm test` 382 passed / 1 skipped. Merging all three PRs in the order above was rehearsed end to end: **zero conflicts**, and the assembled tree typechecks, lints, and passes 382 tests. Two redundant type assertions (`Navbar`, `Sidebar`) were removed to satisfy `no-unnecessary-type-assertion`.

### DIFF
--- a/apps/api/src/setup/dto/update-setup-state.dto.ts
+++ b/apps/api/src/setup/dto/update-setup-state.dto.ts
@@ -6,6 +6,9 @@ import type { BrandingConfig, UpdateSetupStateData } from '@opendatacapture/sche
 @ValidationSchema($UpdateSetupStateData)
 export class UpdateSetupStateDto implements UpdateSetupStateData {
   @ApiProperty({ required: false })
+  activeLanguages?: string[];
+
+  @ApiProperty({ required: false })
   branding?: BrandingConfig | null;
 
   @ApiProperty({ required: false })
--- a/apps/gateway/src/Root.tsx
+++ b/apps/gateway/src/Root.tsx
@@ -1,7 +1,8 @@
 import { useEffect, useRef, useState } from 'react';
 
 import { LanguageToggle, ThemeToggle } from '@douglasneuroinformatics/libui/components';
-import { useNotificationsStore } from '@douglasneuroinformatics/libui/hooks';
+import { useNotificationsStore, useTranslation } from '@douglasneuroinformatics/libui/hooks';
+import type { Language } from '@douglasneuroinformatics/libui/i18n';
 import { CoreProvider } from '@douglasneuroinformatics/libui/providers';
 import { Branding, InstrumentRenderer } from '@opendatacapture/react-core';
 import type { InstrumentSubmitHandler } from '@opendatacapture/react-core';
@@ -14,21 +15,30 @@ import CapWidget from './components/Cap';
 import './services/axios';
 import './services/i18n';
 
+const GATEWAY_LANGUAGES: { [key: string]: string } = { en: 'English', fr: 'Français' };
+const VALID_LANGUAGES = ['en', 'es', 'fr'];
+
 export type RootProps = {
   id: string;
   initialSeriesIndex?: number;
+  lang?: string;
   target: InstrumentBundleContainer;
   token: string;
 };
 
-export const Root = ({ id, initialSeriesIndex, target, token }: RootProps) => {
+export const Root = ({ id, initialSeriesIndex, lang, target, token }: RootProps) => {
   const ref = useRef<HTMLDivElement>(null);
   const notifications = useNotificationsStore();
+  const { changeLanguage } = useTranslation();
 
   const [verified, setVerified] = useState(false);
 
   useEffect(() => {
     ref.current!.style.display = 'flex';
+    const targetLang = lang ?? new URLSearchParams(window.location.search).get('lang') ?? undefined;
+    if (targetLang && VALID_LANGUAGES.includes(targetLang)) {
+      changeLanguage(targetLang as Language);
+    }
   }, []);
 
   // Solving the Cap challenge redeems a short-lived token; exchange it immediately for a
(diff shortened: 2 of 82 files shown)


gdevenyi:
This PR should be last in the stack.

Sorry...

thomasbeaudry:
@gdevenyi You're right that last is mechanically possible — these three branches are independent, each cuts from `main`, and each compiles standalone. So the stack isn't forced by a build dependency. But I'd argue for first, for two reasons.

**1. #1439 emits `?lang=` links that only this PR knows how to read.**

The mailer builds assignment URLs as:

```ts
url: `${assignment.url}?lang=${language}`   // apps/api/src/assignments/assignments.controller.ts
```

Every `?lang` consumer lives in this PR — `apps/gateway/src/{Root.tsx,entry-client.tsx,entry-server.tsx,routers/root.router.ts,services/i18n.ts}`. #1439 touches no gateway files at all.

If the mailer lands first, that's a window where a participant gets a correctly-translated French email whose link opens the instrument in English, because the gateway drops the param. It's not a build break, so CI won't catch it — it's just quietly wrong for anyone who sends a remote assignment in that window.

**2. Going last doesn't avoid the Spanish work, it relocates it into this PR.**

Declaring `es` in `LanguageOptions` makes `es` a *required* key in every `t({ ... })` call in `apps/web` — that's why this PR touches 60 files. The consequence is the same in both orders: whichever of {this, #1439} lands second has to supply Spanish for the other's strings.

- **This first:** #1439 rebases and applies a 183-line backfill. It's already written and verified to apply cleanly, and the Spanish for the mail UI gets reviewed alongside the mail UI itself.
- **This last:** the same 183 lines have to move *into this PR* instead — meaning this stops being a mechanical translation pass over existing code and becomes a PR that also translates the mail UI, which isn't in it. And that patch can't be written until #1439 has actually merged, so the work serializes regardless.

**#1440 is genuinely order-independent** — it adds zero new `t()` calls, so nothing in it changes whichever way this goes. Happy to put it anywhere.

I did rehearse the full merge in the order proposed: exactly one conflict (`SaveStatus.tsx`, add/add with #1439 — take this PR's copy), and the reassembled tree typechecks, lints, and passes 382 tests. If you'd still rather this go last I'll fold the backfill in here instead — it's the same content either way, just a bigger diff on this PR and a `?lang` gap in between.

gdevenyi:
Adding a third language (that none of us reads/writes) is a larger platform discussion than just the technicals.

If there's a technical issue mixed in here (the assignment URLs) split it out.

thomasbeaudry:
Reordered as you asked — this is now last in the stack, and #1440 (UX, fully independent and the smallest) goes first.

One consequence worth flagging: to go last, this PR had to be **stacked on `split/mailer` rather than sitting beside it**. Because declaring `es` makes it required in every `t()` call, this PR now also carries the Spanish for the mail UI — 117 of its lines translate the mail admin page, group email templates, and assignment email form from #1439. That's only possible with those files on the branch, so its base is `split/mailer`; GitHub will retarget it to `main` once #1439 merges.

Two things improved by your ordering, for what it's worth: #1439 no longer needs a follow-up patch, and the `activeLanguages` / `ALL_LANGUAGES` plumbing that was duplicated across both branches now lives in #1439 alone and is inherited here.

The only thing that doesn't get fixed by reordering is the `?lang` gap — #1439 emails links with `?lang=`, and the gateway code reading it is in this PR, so between those two merges an emailed link opens in the gateway's default language. The email body itself is correctly localized; only the linked page is affected. Noted in #1439's description.

Rehearsed the full order (#1440 → #1439 → #1438): zero conflicts, and the assembled tree typechecks, lints, and passes 382 tests.

joshunrau:
This is not done properly. The agent does not understand how libui works. Here are the steps you need to do:

1. Download libui on your machine and make a PR to add optional Spanish support. Also, tell it to document how the translation system works (tell it to refer to that for step 2).
2. Create a new PR that updates libui to this version, and implements the mechanism for Spanish support (do not include any translations. Partial translations are currently accepted.
3. Create a new PR adding the spanish translations (should require no review)

Issue #1086, opened 2025-01-23
add internal edition to card component


CMonnin:
### Current Implementation and Limitations

need to look a schema

### Associated Application Components

Client

### Proposed Solution

change $BaseInstrument to have the edition 

### Estimated Difficulty

None

### Priority

Medium

PR #1487, opened 2026-08-03
Add edition to manage form preview card

david-roper:
closes issue #1086 

Screenshot 2026-08-03 at 4 06 28 PM

### DIFF
--- a/apps/web/src/routes/_app/group/manage.tsx
+++ b/apps/web/src/routes/_app/group/manage.tsx
@@ -306,6 +306,12 @@ const InstrumentPreviewDialog = ({
                   <span className="text-muted-foreground">{item.description}</span>
                 </div>
               )}
+              {item.internal?.edition && (
+                <div>
+                  <span className="font-medium">{t({ en: 'Edition', fr: 'Édition' })}: </span>
+                  <span className="text-muted-foreground">{item.internal.edition}</span>
+                </div>
+              )}
               <div>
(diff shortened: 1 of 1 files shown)


gdevenyi (changes requested):
Thanks — this looks right. The `item.internal` guard matches how the page builds its items (`internal` is null exactly for series), so the row shows for forms and interactive instruments and stays hidden for series, and the label goes through `t()` as required. Nice catch replacing the first commit's `internal?.edition &&` with the plain null check.

One thing before this can merge: nothing exercises the new row. The existing e2e test in `testing/src/specs/group-manage.spec.ts` ('should let a group manager preview an accessible instrument') already opens this dialog, so please:

1. Add `data-testid="preview-edition"` to the new row's `div` in `apps/web/src/routes/_app/group/manage.tsx` — the repo convention (`apps/web/AGENTS.md`) is a testid for any UI an e2e test drives, and this row has no role to select by.
2. In that e2e test, after the dialog assertions, add: `await expect(dialog.getByTestId('preview-edition')).toBeVisible();` — this is deterministic, because the Forms section renders first, so the first "Preview instrument" button always opens a scalar instrument, which always has an edition.

If you hand this to Claude Code, **Sonnet 5** is the right size — it is a testid plus one assertion added to an existing spec, mirroring the surrounding test.

Reviewed at commit 4e95652c93454cefcec1806fb86c2e22d5af8b94.

joshunrau (approved):

Issue #1489, opened 2026-08-04
username not attaching to entry for datahub export


david-roper:
### What Happened?

when a user exports the datahub data all entry in the username section result in N/A when they should have a username attached to them.

### What Did You Expect?

the username of the user who logged the data should be attached to the data entry and should be visible in the datahub export

### Operating System

Ubuntu 22.04

### Browser (if applicable)

Firefox

### Steps to Reproduce

Log in to a user of an odc instance
traverse to datahub containing data
press export and select an file type option
view exported files username column

gdevenyi:
Closed by #1490

PR #1490, opened 2026-08-04
Properly attaching username to data within datahub export

david-roper:
# fix(api): resolve session username in instrument records export

## Summary

- The datahub's "Export" flow (`GET /instrument-records/export`) always showed `N/A` for every
  record's username, regardless of who actually logged the data. `queryRecordsRaw`'s aggregation
  pipeline joined the `Session` collection but then read `session.user.username` in its `$project`
  stage — `Session` documents never embed a `user` object, they only store a `userId` reference, so
  that field was always empty.
- Fix the pipeline to actually resolve the user: cast `session.userId` to a proper `ObjectId` via
  `$convert` (this also normalizes a small number of legacy sessions whose `userId` had been
  persisted as a plain string instead of an `ObjectId`), `$lookup` against `UserModel` on the cast
  field, and project the username from the joined document.
- Add a regression test asserting the pipeline handed to `aggregateRaw` contains the `UserModel`
  lookup and projects `$sessionUser.username`. The existing export test mocked `aggregateRaw`'s
  return value directly, so it exercised the worker's row-mapping logic but could never have caught
  a broken aggregation pipeline shape — which is exactly how this bug shipped silently.

## Why this was invisible until now

`aggregateRaw` bypasses Prisma's typed query builder, so neither `tsc` nor a schema check could
catch the mismatch between the assumed `session.user` shape and the actual `Session` model. The
inline comment removed by this PR (`// TBD test this works`) was a pre-existing admission that this
exact line had never been verified.

## Test plan

- [x] `pnpm exec vitest --project api` — all tests pass, including the new regression test
- [x] `pnpm lint` / `pnpm test` from the repo root
- [x] Replayed the new pipeline shape directly against a full copy of production-scale data: went
      from 0 resolved usernames on any exported record to correctly resolving the username on every
      record whose session actually has one attached, including the legacy string-typed `userId`
      cases

closes issue #1489 

assisted with claude  Sonnet 5

### DIFF
--- a/apps/api/src/instrument-records/__tests__/instrument-records.service.spec.ts
+++ b/apps/api/src/instrument-records/__tests__/instrument-records.service.spec.ts
@@ -419,5 +419,22 @@ describe('InstrumentRecordsService', () => {
         value: 85
       });
     });
+
+    it('should join and project the session user by a cast-hardened lookup, so the export includes the username of whoever ran the session', async () => {
+      instrumentRecordModel.aggregateRaw.mockResolvedValueOnce([]);
+      const ability = { can: () => true } as any;
+
+      await instrumentRecordsService.exportRecords({}, { ability });
+
+      const [{ pipeline }] = instrumentRecordModel.aggregateRaw.mock.lastCall as [{ pipeline: any[] }];
+
+      // Session documents only ever store a userId reference, never an embedded user object, so
+      // resolving a username requires an explicit join against UserModel.
+      const userLookupStage = pipeline.find((stage) => stage.$lookup?.from === 'UserModel');
+      expect(userLookupStage).toBeDefined();
+
+      const projectStage = pipeline.find((stage) => stage.$project);
+      expect(projectStage.$project.session.user.username).toBe('$sessionUser.username');
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


gdevenyi (changes requested):
Thanks — the diagnosis is exactly right, and the pipeline change checks out against the schema (`User` → `UserModel` collection, `Session.userId` as an optional ObjectId), including the graceful `N/A` fallback for sessions without a user. Two things before this can land:

1. Please add an e2e test in `testing/src/specs/datahub.spec.ts` (see `.agents/docs/playbooks/add-e2e-test.md`): complete an instrument as the logged-in user, run the datahub export, and assert the CSV shows that username instead of `N/A`. Every change here needs both a unit test and an e2e test, and in this case the e2e is the only test in the repo that executes this aggregation against a real database — `aggregateRaw` is invisible to the type checker and the unit suite mocks the DB, which is how the original bug shipped.
2. In the new regression test (`apps/api/src/instrument-records/__tests__/instrument-records.service.spec.ts`, around line 433), also assert that the `UserModel` lookup stage has `as: 'sessionUser'` (and ideally `localField: 'session.userObjectId'`). Right now, renaming the lookup alias while the projection still reads `$sessionUser.username` would reintroduce exactly the bug this PR fixes — a projection referencing a field no stage produces — and the test would still pass.

If you hand this to Claude Code, **Opus 5** is the right size — the e2e spans testing fixtures, the datahub page flow, and a download assertion, which needs more repo judgment than copying a single existing test.

Reviewed at commit c76541215d0eb81e757ebf5056dd2f30582864e4.

PR #1497, opened 2026-08-05
feat(i18n): translate the remaining interface strings into Spanish

thomasbeaudry:
## Summary

#1442 made Español selectable and translated the 11 `apps/web` namespace JSON files. It did not touch the **inline** `t()` calls, and there were 572 of them naming only `en` and `fr`.

libui resolves `obj[resolvedLanguage] ?? obj.en`, so every one of those rendered in English for a user reading in Spanish, silently. The admin panel and its submenu — reported by @thomasbeaudry — were among them:

```ts
label: t({ en: 'Admin Panel', fr: "Panneau d'administration" })   // → "Admin Panel" in Español
```

This adds the missing `es` entry at all 572 sites.

## Scope

| Workspace | Sites |
| --- | --- |
| `apps/web` | 503 |
| `packages/react-core` | 63 |
| `apps/gateway` | 6 |

**`apps/playground` is deliberately excluded** (107 more sites). `Header.tsx` and `IndexPage.tsx` hard-code their `LanguageToggle` options as `{ en, fr }` — it uses libui's toggle directly rather than react-core's `activeLanguages`-driven one — so Spanish cannot be selected there at all. Translating it means widening that toggle first, which is a behaviour change to a second app and belongs in its own PR.

## Terminology

The completed namespace files already establish Spanish vocabulary, so the inline strings reuse it rather than inventing a parallel one: **Panel de control** (Dashboard), **Centro de datos** (Data Hub), **Sujeto** (Subject), **Tarea remota** (Remote Assignment), **Iniciar una sesión** (Start Session), **Administrar instrumento**, **Comuníquese con el administrador de la plataforma**. Register is *usted* throughout, matching the existing files.

Three strings needed resolving per call site rather than globally, because English collapses distinctions Spanish does not:

- `'Subject'` → **Asunto** in `EmailTemplateEditor` (an email's subject), **Sujeto** in `admin/users` and `InstrumentSummary` (the participant — the existing `fr` there is `Client`).
- `'Default'` → **Predeterminada** on a template row, **Predeterminado** on a font size.

Five strings carry no words: the `'{}: {}'` / `': '` format strings whose `fr` variant exists only because French puts a space before a colon. Spanish follows the English convention, so they get an entry identical to `en` — explicit rather than left half-deliberate, half-forgotten.

## Keeping it

Translating 572 strings once is worth little if the 573rd ships untranslated next week. `apps/web/src/__tests__/spanish-coverage.test.ts` parses every `t({…})` in all three frontends and fails on one that names `en` but not `es`, reporting the file and the call:

```
apps/web/src/hooks/useNavItems.ts — { en: 'Mail', fr: 'Courriel' }
```

It lives in `apps/web` because that is the only one of the three frontends with a vitest project. The scanner is quote- and comment-aware — a brace inside a string is text, not structure — and two of its three cases assert that directly, so the guard cannot pass by failing to parse.

The four `AGENTS.md` files that taught the `{ en, fr }` shape now teach `{ en, es, fr }` and point at the test, so a contributor learns the rule before the build tells them. (`packages/react-core/AGENTS.md` also still described `LocalizedText` as `{ en?, fr? }`; it has been keyed by `Language` since #1442.)

## Also

`Dropzone`'s `'Invalid Input'` and `UploadProgressBar`'s upload counter were English-only before this — adding `es` alone would have left them conspicuously `{ en, es }` with no French, so both got their missing `fr` too.

## Test plan

- [x] `pnpm test` — 657 passed, 1 skipped (3 new)
- [x] `pnpm test:e2e` — 142 passed, full suite, forced past turbo's cache

The e2e in `admin-settings.spec.ts` activates Spanish, switches the toggle to Español, and asserts the nav group reads **Panel de administración** and its submenu **Configuración de la aplicación** — the exact reported case. It extends the existing language test rather than adding a new one: `activeLanguages` is a single instance-wide document and `fullyParallel` is on, so a second test mutating it would race the "only one language left" assertion.

Both new tests were verified to fail when the translations are removed, rather than assumed to be checking anything.


apps/web/src/__tests__/spanish-coverage.test.ts
```diff
@@ -0,0 +1,99 @@
+import { readdirSync, readFileSync, statSync } from 'node:fs';
+import { join, relative, resolve } from 'node:path';
+
+import { describe, expect, it } from 'vitest';
+
+const REPO_ROOT = resolve(import.meta.dirname, '../../../..');
+
+/**
```
joshunrau: this test should not be required. can you update libui and then use the new option to require full translations at the type levle

joshunrau (changes requested):

gdevenyi (changes requested):
Thanks — this is careful work: the terminology is consistent across all 517 strings, the type guard is the right replacement for a runtime coverage test, and proving it fails when a key is removed checks out. A few items before this can land:

1. The upload error page regressed: `t({ en: d.en ?? '', es: d.es ?? '', fr: d.fr ?? '' })` in `apps/web/src/routes/_app/upload/$instrumentId.tsx` renders an empty description for Spanish readers (plus a `console.error`), because every `UploadError` in `apps/web/src/utils/upload.ts` still carries only `en`/`fr` and an empty string defeats libui's nullish fallback. Please restore the English fallback (`es: d.es ?? d.en ?? ''`, same for `fr`) or translate the 41 descriptions in `upload.ts` — and add a unit test for this path, since the type guard structurally cannot see it and the PR has no unit test otherwise.
2. In `SeriesInstrumentRenderer.tsx:238`, `es: completionMessage?.en ?? …` is correct but looks like a typo — a one-line comment will stop someone "fixing" it later.
3. `packages/react-core/AGENTS.md:42` still calls `LocalizedText` "a partial record"; it is total now.
4. Three docs still teach `t({ en, fr })`, which now fails `tsc` in the workspaces you covered: `.agents/docs/playbooks/promote-to-react-core.md:41`, `.agents/docs/playbooks/add-web-data-hook.md:57`, and `.agents/docs/packages/libui.md:41` (which should also mention `requireCompleteTranslations`).
5. The login and setup pages (`apps/web/src/routes/auth/login.tsx`, `apps/web/src/routes/setup.tsx`) hardcode `{ en, fr }` toggle options, so an admin can activate Spanish but nobody can select it before logging in, and the setup wizard cannot run in Spanish. react-core's `activeLanguages`-driven `LanguageToggle` looks like the drop-in fix; otherwise document the exclusion the same way you documented playground's.
6. Two small e2e items in `testing/src/specs/admin-settings.spec.ts`: the title at line 51 no longer describes a body that now also asserts Spanish nav rendering, and line 82 should use the `nav-button-/admin/settings` testid instead of role+name (`testing/AGENTS.md` reserves roles for elements with no testid).
7. Two minor Spanish fixes: `InstrumentSummary.tsx:95` — `Resumen de resultados del ${title}` forces the masculine `del` before any title; reorder as the French does. `SeriesInstrumentContent.tsx:46-48` — the `/{${…}}` template renders a literal `3/{5}`; it predates your PR in en/fr, but since the new es copies it, this is the moment to fix all three.

If you hand this to Claude Code, **Fable 5** is the right size — the fixes span runtime behavior, react-core, e2e and docs, and the PR touches the workspace catalog.

Reviewed at commit 96e9f1d08b2470bb2ee6439c95e69987f3f896e5.

joshunrau (changes requested):
This branch no longer merges cleanly into `main` — GitHub reports the merge as conflicting, so the merge result cannot be built or reviewed. CI itself is green.

Please update the branch against current `main` and resolve the conflicts. Review resumes once it is mergeable.

Reviewed at commit 5ae9241ce1d3b07e08016c2774f655f753e58070.

thomasbeaudry:
Updated against current `main` in c61f9c6a5. The branch merges cleanly again.

**Conflicts resolved**
- `pnpm-workspace.yaml` / `pnpm-lock.yaml`: took `main`'s libui `^6.16.1` and libui-form-types `^1.4.0`, then reinstalled. `requireCompleteTranslations` still works on 6.16.1, and the lockfile is identical to `main`'s.
- `WalkthroughProvider.tsx`, `admin/settings.tsx`: kept `main`'s `isGatewayEnabled` gating and put the `es` strings back inside it.
- `admin/users/index.tsx`: kept `main`'s `common.groupRequired` key and the extracted `additionalPermissionsFieldset`, and added the `es` entries to that fieldset (same terms as before).
- `AGENTS.md`: kept `main`'s new test-exception and Conventional Commits bullets, plus the translation rule from this branch.

**Strings `main` added since the last review, which the type guard now rejected**
- `admin/audit/logs.tsx`: five filter labels.
- `auth/reset-password.tsx`: six strings. This page also hardcoded an `{ en, fr }` language toggle, so it now uses react-core's `activeLanguages`-driven toggle like login and setup (same reasoning as review item 5).

The Spanish additions ended up in the merge commit itself rather than a separate commit.

All of @gdevenyi's review items and @joshunrau's inline comment on the coverage test were already addressed in 5ae9241ce.

**Checks**
- `lint`: web, gateway, react-core and testing pass.
- Unit tests (web, react-core, gateway): 372/372. `gateway-route-guards` timed out once while running concurrently with lint and passes on its own.
- e2e `admin-audit-logs`, `admin-management`, `admin-settings`, `password-reset` (chromium): 37/37.

Issue #1494, opened 2026-08-05
make group mandatory for non admin users


david-roper:
### Current Implementation and Limitations

current form allows for any type of user to be created by an admin and not have a group. whilst the update user form requires them to have one, thus making it inconsistent. 

### Associated Application Components

Client

### Proposed Solution

make it so in both forms having a group is mandatory unless that user is an admin.
adjust the schema logic of both forms to allow this behaviour to be possible.

### Estimated Difficulty

Low

### Priority

Low

Issue #1472, opened 2026-07-27
Admin 'manage user' edit sheet silently requires a group, asymmetric with the create form


joshunrau:
## Summary

The admin "Manage" sheet for a user cannot save any edit unless the user has at least one group — and the create-user form allows creating a user with none. The result is a user who can be created but never edited, with no visible reason.

## Where

- `apps/web/src/routes/_app/admin/users/index.tsx` — the update schema `.check()` pushes an inline "Standard user must be part of a group" issue whenever `groupIds` is empty for any role other than `ADMIN`.
- `apps/web/src/routes/_app/admin/users/create.tsx` — `groupIds` is fully optional.

The two forms disagree, so a `GROUP_MANAGER`/`STANDARD` user created with no groups is permanently unsavable from the Manage sheet.

## Repro

1. Create a user via the API (or the create form) with no `groupIds`
2. Open `/admin/users` → row actions → "Manage"
3. Change only the Email field, click Submit

Result: no request fires, no toast, and no top-level error. The only signal is an inline field error next to "Groups", which is off-screen unless you scroll.

## Suggested fix

Make the two schemas agree — either require a group at creation for non-admin roles, or drop the requirement on update. Whichever way, surface the failure at the top of the sheet rather than only inline, since the submit otherwise looks like a no-op.

## Notes

Found while writing the Playwright suite (branch `e2e-tests`). `testing/src/specs/admin-management.spec.ts` seeds its user *with* a group to work around this.

PR #1498, opened 2026-08-05
User group manadatory

david-roper:
# feat(web): require a group for non-admin users in the admin user forms

## Summary

A non-admin user with no group is a broken account: every read rule granted to a
`GROUP_MANAGER` or `STANDARD` user is scoped to the groups they belong to, so a groupless
one can still create sessions, subjects and records — but nothing they produce is readable
by any group manager, only by an admin. `InstrumentRecordsService.upload` already rejects
such users outright ("The Non-Admin User must be part of a group"), so the app was happily
creating accounts guaranteed to fail later.

The update form already enforced this rule; the create form did not (the asymmetry noted in
#1472). This closes the gap and makes both forms enforce one shared rule:

| User                                    | Group required? |
| --------------------------------------- | --------------- |
| `ADMIN`                                 | No              |
| `GROUP_MANAGER` / `STANDARD`            | **Yes**         |
| `basePermissionLevel: null`             | **Yes**         |
| Any user marked *Disabled*              | No              |

Disabled accounts are exempt because they exist only to attribute uploaded data, never to
log in and read it.

## Implementation

- Adds `requiresGroup({ basePermissionLevel, disabled })` to `apps/web/src/utils/validation.ts`.
  Both forms call it, so the create/update pair cannot drift apart again, and the rationale
  lives in exactly one place.
- Wires the predicate into the create form's validation schema (new behaviour) and into the
  update form's existing check (which gains the disabled exemption).
- Moves the message to `common.groupRequired` (en/es/fr) now that it is used twice, and
  rewords it: the old string said *"Standard user must be part of a group"*, which was wrong —
  the rule also applies to group managers.

### Two defects fixed in the update form along the way

- **Stale validation schema.** The `useMemo` building the schema closed over
  `data.selectedUserBasePermission` but did not list it as a dependency. Opening an admin and
  then a non-admin whose phone number was also blank reused the schema built for the *previous*
  user, validating against the wrong permission level. Now a dependency.
- `input: ctx.value.confirmPassword` on an issue whose `path` is `['groupIds']` → `ctx.value.groupIds`.

## Scope

Deliberately client-side only. The API and shared schemas are unchanged, so a groupless
non-admin can still be created over the API or `cli/odc-cli`. Enforcing it at the boundary
was considered and consciously deferred — it changes `$CreateUserData`, `UsersService`, and
the e2e `createUser` helper's defaults, and is better as its own PR.

This is a data-governance fix, not a security fix: a groupless non-admin **fails closed**
(every read filter resolves to an empty set), so there is no privilege escalation in either
the old or new behaviour.

## Test plan

- [x] `pnpm lint` — 33/33 tasks
- [x] `pnpm test` — 657 passed, 1 skipped
- [x] `pnpm --filter @opendatacapture/testing test:e2e src/specs/admin-management.spec.ts` — 19/19
- [x] Unit: 6 cases in `validation.test.ts` covering admin, both non-admin levels, null, and
      disabled. Verified they fail when `requiresGroup` is stubbed out (4 red), then restored.
- [x] E2E: rejection of a groupless non-admin, and successful creation of a *disabled*
      groupless one. Verified the rejection test fails with the rule disabled.

> Note: `gateway-assignment … report it back to the clinician @smoke` fails on firefox in my
> environment, but it fails identically on unmodified `main` — pre-existing and unrelated.

Closes issue #1494 
Closes issue #1472 
model sonnet 5

### DIFF
--- a/apps/web/src/routes/_app/admin/users/create.tsx
+++ b/apps/web/src/routes/_app/admin/users/create.tsx
@@ -18,7 +18,7 @@ import { useCreateUserMutation } from '@/hooks/useCreateUserMutation';
 import { groupsQueryOptions, useGroupsQuery } from '@/hooks/useGroupsQuery';
 import { useMailErrorMessage } from '@/hooks/useMailErrorMessage';
 import { useSetupStateQuery } from '@/hooks/useSetupStateQuery';
-import { $Email, $PhoneNumber, omittedIfBlank } from '@/utils/validation';
+import { $Email, $PhoneNumber, omittedIfBlank, requiresGroup } from '@/utils/validation';
 
 const PASSWORD_ERROR_TRANSLATION_KEYS = {
   INSUFFICIENT_PASSWORD_STRENGTH: 'common.insufficientPasswordStrength',
@@ -309,6 +309,16 @@ const RouteComponent = () => {
                 input: ctx.value.confirmPassword,
                 message: t('common.passwordsMustMatch'),
                 path: ['confirmPassword']
+              });
+            }
+          })
+          .check((ctx) => {
+            if (requiresGroup(ctx.value) && !ctx.value.groupIds?.size) {
+              ctx.issues.push({
+                code: 'custom',
+                input: ctx.value.groupIds,
+                message: t('common.groupRequired'),
+                path: ['groupIds']
               });
             }
           })}
--- a/apps/web/src/routes/_app/admin/users/index.tsx
+++ b/apps/web/src/routes/_app/admin/users/index.tsx
@@ -19,7 +19,7 @@ import { groupsQueryOptions, useGroupsQuery } from '@/hooks/useGroupsQuery';
 import { useUpdateUserMutation } from '@/hooks/useUpdateUserMutation';
 import { usersQueryOptions, useUsersQuery } from '@/hooks/useUsersQuery';
 import { useAppStore } from '@/store';
-import { $Email, $PhoneNumber, clearedIfBlank, omittedIfUnchanged } from '@/utils/validation';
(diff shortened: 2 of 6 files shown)


gdevenyi (changes requested):
Thanks — the shared `requiresGroup` predicate, the create-form gap closure, and the two update-form fixes (stale `useMemo` schema, wrong `input` on the groupIds issue) all look right, and the create-path tests are solid.

One real gap on the update form:

1. The disabled exemption never fires for a user who is already disabled. The `useEffect` that builds the form input (`apps/web/src/routes/_app/admin/users/index.tsx:355-371`) never puts `disabled` into `initialValues`, so the radio always starts at `false`. Open the manage sheet for a disabled, groupless account — exactly what the create form now permits — and saving is blocked by the new group-required error; add a group and save, and the account is silently re-enabled because `disabled: false` is sent. Please add `disabled: selectedUser.disabled ?? false` to both `initialValues` branches. The omission predates this PR, but your new check on `ctx.value.disabled` makes it matter.
2. Please add an update-path e2e test in `testing/src/specs/admin-management.spec.ts`: seed `api.createUser({ disabled: true })` (the helper already defaults `groupIds: []`), open the manage sheet, submit without changes, and assert the save succeeds and the account stays disabled. Modelled on the existing manage-sheet test, this is the test that fails today under the bug in item 1 — the update form's changed validation currently has no test at all.

If you hand this to Claude Code, **Opus 5** is the right size — the fix spans the stored value, the form's initial values, the exemption, and the save payload, and the test has to assert that chain end to end.

Reviewed at commit 8582848da5ed88568944bcd82271f39eeda25f88.

joshunrau (approved):

PR #1502, opened 2026-08-06
feat(settings): let an instance turn off remote assignments

thomasbeaudry:
## What

Adds **Enable Remote Assignments** to the Features section of App Settings, directly under Enable
Uploader, with a `?` hover card explaining what it does. It is **enabled by default** — including on
every instance set up before the setting existed.

Instances that do not use the gateway never need remote assignments, so turning this off removes:

- the **Remote Assignment** item from the session nav in the sidebar
- the **Assignments** tab on a subject in the datahub

## How

- `packages/schemas`: `isRemoteAssignmentsEnabled` on `$SetupState` (a plain `boolean`, never null —
  the API resolves an unset value) and on `$UpdateSetupStateData` (optional).
- `apps/api`: `Boolean?` column on `SetupState`; `getState` resolves an unsaved value to `true`, so
  the opt-out defaults to off. `PATCH /v1/setup` already requires `manage all`.
- `apps/web`: `useIsRemoteAssignmentsEnabled` answers "does this instance offer remote assignments?"
  once — gateway deployed *and* feature on — and both the sidebar and the subject tab read it. The
  datahub tab previously read the build-time `config.setup.isGatewayEnabled` while the sidebar read
  the API's value; both now use the API's.
- The settings page's repeated label + help + toggle row is factored into one `FeatureToggle`, and
  the duplicated hover-card markup into `HelpHoverCard`.

## Tests

- `apps/web/src/hooks/__tests__/useIsRemoteAssignmentsEnabled.test.ts` — the full gateway × feature
  matrix.
- `apps/web/src/hooks/__tests__/useNavItems.test.ts` — the nav item appears only when both hold.
- `apps/api/src/setup/__tests__/setup.service.spec.ts` — unset reads as enabled, saved `false` reads
  as disabled, and the flag persists.
- `packages/schemas/src/setup/setup.test.ts` — the update field accepts booleans, may be omitted, and
  rejects `null`.
- `testing/src/specs/admin-settings.spec.ts` — toggling the setting adds and removes the sidebar item.
  It is the one spec that touches this instance-wide flag and it turns it back on before finishing,
  because the rest of the suite drives remote assignments.

`pnpm lint`, `pnpm test` (668 passed) and `pnpm test:e2e` (143 passed) are green.

## Note

Disabling the feature hides the UI; the `/session/remote-assignment` and
`/datahub/$id/assignments` routes and the API endpoints remain reachable by direct URL, matching how
the Enable Uploader flag behaves today.

### DIFF
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -439,6 +439,7 @@ model SetupState {
   defaultAssignmentDurationDays Int?
   isDemo                        Boolean
   isExperimentalFeaturesEnabled Boolean?
+  isRemoteAssignmentsEnabled    Boolean?
   isSetup                       Boolean
   mailConfig                    MailConfig?
   newUserEmailTemplate          MailTemplate?
--- a/apps/api/src/setup/__tests__/setup.service.spec.ts
+++ b/apps/api/src/setup/__tests__/setup.service.spec.ts
@@ -43,6 +43,15 @@ describe('SetupService', () => {
         where: { id: 'setup-1' }
       });
     });
+
+    it('should persist isRemoteAssignmentsEnabled', async () => {
+      setupStateModel.findFirst.mockResolvedValue({ id: 'setup-1', isSetup: true });
+      await setupService.updateState({ isRemoteAssignmentsEnabled: false });
+      expect(setupStateModel.update.mock.lastCall?.[0]).toMatchObject({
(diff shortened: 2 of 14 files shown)


joshunrau:
This is completely wrong. Right now, this is stored in the env. The entire backend architecture is based on this env variable with conditional module loading. Just use the env variable in the frontend. Pass this into the config on frontend and use that.

Issue #1411, opened 2026-07-20
Performance: /v1/audit/logs returns the entire audit log with no pagination, limit or sort


gdevenyi:
### Current Implementation and Limitations

`AuditService.find()` (`apps/api/src/audit/audit.service.ts:10`) returns **every** matching audit log row, with no `take`, no `skip`, no cursor and no `orderBy`:

```ts
return this.auditLogModel.findMany({
  select: { action: true, entity: true, group: { select: { name: true } }, id: true, timestamp: true, user: { select: { username: true } } },
  where: { action, entity, groupId, userId }
});
```

When the admin audit log page is opened with no filters selected, `where` is `{ action: undefined, entity: undefined, groupId: undefined, userId: undefined }` — every row in the collection.

The client mirrors this. `useAuditLogsQuery` (`apps/web/src/hooks/useAuditLogsQuery.ts:8`) fetches the whole array, validates it with `z.array($AuditLog).parse(...)`, and sorts it in JavaScript:

```ts
return z.array($AuditLog).parse(response.data).sort((a, b) => b.timestamp - a.timestamp);
```

and `routes/_app/admin/audit/logs.tsx:287` hands the entire array to `<DataTable>`, which paginates client-side — so every row is held in memory and in the table's row model regardless of how few are visible.

The distinguishing property of this collection is that **it only ever grows**. `AuditLogger` records every `CREATE`, `UPDATE`, `DELETE` and every `LOGIN`. There is no retention policy, no archival, and nothing bounds it. An instance that has been running for a year with regular use will have a table that is orders of magnitude larger than anything else the UI renders in one page.

There is no index on `AuditLog` either (`timestamp`, `groupId`, `userId` are all unindexed), so every load is a full collection scan and the sort cannot be served by an index — see the separate indexing issue.

Two smaller issues in the same file:

- `logs.tsx:140` — `handleDownload` is wrapped in `useCallback` with an empty dependency array but closes over `table`, so it captures the first render's table. Not a performance bug, but it will silently export stale rows.
- `logs.tsx:147` — `users.filter((user) => user.groupIds.includes(search.groupId!))` recomputes on every render of `Toggles`; trivial at current user counts but easily memoised.

### Measurements

120,000 audit log rows (a plausible year of activity for a modestly-used instance):

```
auditlogs {}         120,000 docs examined -> 120,000 returned, 17 ms   COLLSCAN
[...]

Issue #1409, opened 2026-07-20
Performance: dashboard summary issues 20 sequential count queries and loads every instrument


gdevenyi:
### Current Implementation and Limitations

`SummaryService.getSummary()` (`apps/api/src/summary/summary.service.ts:23`) backs the dashboard — the first page a user sees after logging in. It issues **20 database round trips, every one of them `await`ed sequentially**, even though all 20 are independent:

```ts
const counts = {
  instruments: await this.instrumentsService.count(),
  records:     await this.instrumentRecordsService.count(recordFilter, { ability }),
  sessions:    await this.sessionsService.count(sessionFilter, { ability }),
  subjects:    await this.subjectsService.count(filter, { ability }),
  users:       await this.usersService.count(filter, { ability })
};

for (let i = 0; i < 5; i++) {
  const lte = new Date(now - THIRTY_DAYS_MS * i);
  trends.records.push({  timestamp, value: await this.instrumentRecordsService.count({ date: { lte }, ...recordFilter }, { ability }) });
  trends.sessions.push({ timestamp, value: await this.sessionsService.count({ date: { lte }, ...sessionFilter }, { ability }) });
  trends.subjects.push({ timestamp, value: await this.subjectsService.count({ createdAt: { lte }, ...filter }, { ability }) });
}
```

Three problems compound here:

1. **Serialisation.** 20 sequential awaits means total latency is the *sum* of 20 round trips rather than the max. Nothing in the loop depends on a previous iteration.
2. **`instrumentsService.count()` is not a count.** It is implemented as `(await this.find(query, options)).length` (`instruments.service.ts:58-63`), so the dashboard fetches and instantiates every instrument in the database — bundles included — to produce one integer. (Covered in more detail in the separate issue on instrument bundle over-fetching.)
3. **The five trend points are five scans of the same collection.** `trends.records` counts records with `date <= T` for five values of `T` on the same filter; these are five passes over the same data that a single `$bucket`/`$group` aggregation would answer in one.

Without indexes (see the separate indexing issue) each of these counts is a full collection scan, so the cost is 20 × COLLSCAN.

### Measurements

MongoDB 8, 200,000 instrument records / 50,000 sessions / 5,000 subjects, warm cache:

```
20 sequential counts, no indexes    median 203 ms
20 sequential counts, with indexes  median  37 ms
```

[...]

PR #1513, opened 2026-08-25
perf(api): create the mongodb indexes every query depends on

### LINKED ISSUE #1406: Performance: no indexes are ever created on any MongoDB collection

joshunrau:
Closes #1406.

## The problem

MongoDB creates a collection on first insert with only its `_id_` index, and **nothing in `apps/api` ever runs `prisma db push`** — there is no `db:push` script, the Dockerfile only builds, and the only workspace that owns one is `apps/gateway`, whose datasource is SQLite. A freshly provisioned instance therefore has one index per collection, so:

- every query filtering on anything but `_id` is a full collection scan, and
- the `@unique` constraints the schema declares on `Group.name`, `InstrumentRepo.url` and `InstrumentRecord.assignmentId` are enforced by nothing.

## Why not just run `prisma db push`

That is what #1406 proposes, and **it would take the API down.**

`InstrumentRecord.assignmentId` is optional but has to carry `@unique`: Prisma rejects the schema without it (`P1012` — a 1:1 relation needs a unique field on the defining side). `db push` builds that index **non-sparse**, and MongoDB indexes a *missing* field as `null`. So the second instrument record collected outside a remote assignment collides.

Reproduced against the unmodified schema on this branch's parent:

```
push: [+] Unique index `InstrumentRecordModel_assignmentId_key` on ({"assignmentId":1})
record #1 (no assignment): CREATED
record #2 (no assignment): FAILED -> P2002
records stored: 1 | assignmentId present in doc? [ false ]
```

The only reason the app works today is that the indexes were never created. Prisma's MongoDB connector has no `sparse` option, so there is no way to spell a correct index set in `schema.prisma` alone.

## What this does

- **Declares the indexes in `schema.prisma`** (15 `@@index`, plus `@@unique([username])` on `User`, which was missing). `db push` drops any index the schema does not name, so they have to be declared even though the schema is not what creates them.
- **Creates them at startup**, from `ensureDatabaseIndexes` in `src/core/prisma.ts`, which the client factory awaits after `$connect()`. Index names reproduce Prisma's own convention, so a later `db push` reads them as already satisfied rather than making duplicates.
- **Creates the `assignmentId` index sparse, and replaces a non-sparse one it finds** — so an instance that has already been pushed to heals on its next boot. Verified that `db push` then leaves the sparse index alone.
- Fails startup loudly if an index cannot be built (a unique index cannot be built over existing duplicates), naming the collection.

`Assignment.status` is deliberately left unindexed — #1406 proposes it, but it is written and never filtered on.

## Benchmarks

`mongodb-memory-server` (MongoDB 7) single-node replica set, 395,000 documents seeded to match the
schema exactly: 5,000 subjects, 50,000 sessions, 200,000 instrument records, 120,000 audit logs,
20,000 record files, with one "core intake" instrument holding 40% of records. Same process, same
data, warm cache — the **only** variable between the two columns is whether the indexes exist.

`docs examined` and the plan come from `explain('executionStats')`; times are the median of 5 runs of
the same query issued through the Prisma client. A warm cache with the whole working set in RAM is
the *best* case for the unindexed side — once a collection outgrows RAM a `COLLSCAN` goes to disk and
the gap widens.

| Query (the shape the service issues) | examined → returned before | after | ms before | ms after | plan |
|---|---|---|---|---|---|
| records for a subject (datahub) | 200,000 → 18 | 18 → 18 | **191** | **1** | COLLSCAN → IXSCAN |
| records for a subject + instrument | 200,000 → 40 | 40 → 40 | **171** | **2** | COLLSCAN → IXSCAN |
| records for a session | 200,000 → 4 | 4 → 4 | **85** | **<1** | COLLSCAN → IXSCAN |
| audit logs, group + sorted page | 120,000 → 25 | 25 → 25 | **49** | **<1** | COLLSCAN **+ in-memory SORT** → IXSCAN |
| files for a record | 20,000 → 1 | 1 → 1 | **8** | **<1** | COLLSCAN → IXSCAN |
| record count for a group + date (summary) | 200,000 → 40,000 | 40,000 → 40,000 | 95 | 47 | COLLSCAN → IXSCAN |
| records by instrument (`instruments/info`, #1416) | 200,000 → 80,000 | 80,000 → 80,000 | 273 | 218 | COLLSCAN → IXSCAN |
| sessions for a group | 50,000 → 10,000 | 10,000 → 10,000 | 118 | 106 | COLLSCAN → IXSCAN |
| subjects for a group | 5,000 → 1,000 | 1,000 → 1,000 | 12 | 11 | COLLSCAN → IXSCAN |

**Read the table in two halves.** The selective queries — the datahub's own reads, a session's
records, a page of audit logs — go from examining the whole collection to examining exactly what they
return, and the work drops by two orders of magnitude. The bottom four still examine tens of
thousands of documents *because they return* tens of thousands; the index removes the scan but not
the cost of materialising the result, so they improve only modestly. **Indexes are necessary here,
not sufficient** — #1409, #1410 and #1411 are what move those.

The audit-log row is the one that is not merely slow: without an index the sort is done **in memory**
over a full collection scan, and MongoDB aborts a non-indexed sort once it exceeds 32 MB. That
endpoint does not degrade at scale, it starts failing.

### Cost

| Collection | data | index before (`_id_` only) | index after | added | added / data |
|---|---|---|---|---|---|
| InstrumentRecordModel | 50.73 MiB | 4.82 MiB | 11.95 MiB | 7.13 MiB | 14.0% |
| AuditLogModel | 14.02 MiB | 1.93 MiB | 6.35 MiB | 4.41 MiB | 31.5% |
| SessionModel | 8.00 MiB | 1.21 MiB | 2.31 MiB | 1.11 MiB | 13.8% |
| InstrumentRecordFileModel | 2.87 MiB | 0.00 MiB | 0.19 MiB | 0.18 MiB | 6.4% |
| SubjectModel | 0.65 MiB | 0.14 MiB | 0.19 MiB | 0.05 MiB | 7.8% |
| **total** | **76.27 MiB** | | | **12.88 MiB** | **16.9%** |

Startup cost, which is what `ensureDatabaseIndexes` adds to a boot:

- **first boot against this dataset — 1.89s** (builds all 17 indexes)
- **every subsequent boot — 8ms** (everything present; nothing is rebuilt)

<details>
<summary>Benchmark script (not committed — run it from <code>apps/api</code> with <code>pnpm exec tsx</code>)</summary>

```ts
/**
 * Benchmarks the query shapes apps/api services actually issue, with and without the indexes
 * this change creates. Same process, same data, same warm cache — the only variable is the indexes.
 */
import { MongoMemoryReplSet } from 'mongodb-memory-server';
import { MongoClient, ObjectId } from 'mongodb';
import { PrismaClient } from '@prisma/client';
import { ensureDatabaseIndexes } from '../src/core/prisma.ts';

const SUBJECTS = 5_000;
const SESSIONS = 50_000;
const RECORDS = 200_000;
const AUDIT_LOGS = 120_000;
const FILES = 20_000;
const BATCH = 5_000;
const REPEATS = 5;

const rs = await MongoMemoryReplSet.create({ replSet: { count: 1, name: 'rs0' } });
const uri = new URL(rs.getUri('bench')).href;
const mc = new MongoClient(uri);
await mc.connect();
const db = mc.db('bench');
const prisma = new PrismaClient({ datasourceUrl: uri });
await prisma.$connect();

// ---------------------------------------------------------------- fixtures
const groupIds = Array.from({ length: 5 }, () => new ObjectId());
const userIds = Array.from({ length: 10 }, () => new ObjectId());
// One "core intake" instrument holds 40% of records, mirroring how a real deployment skews.
const instrumentIds = ['CORE_INTAKE', 'HAPPINESS', 'DEMOGRAPHICS', 'CONSENT', 'MRI_SESSION'];
const subjectIds = Array.from({ length: SUBJECTS }, (_, i) => `subject-hash-${i}`);
const now = Date.now();
const pick = <T>(xs: T[], i: number): T => xs[i % xs.length]!;
const someDate = (i: number) => new Date(now - (i % 900) * 86_400_000);

const insertAll = async (collection: string, total: number, make: (i: number) => any) => {
  for (let start = 0; start < total; start += BATCH) {
    const size = Math.min(BATCH, total - start);
    await db.collection(collection).insertMany(Array.from({ length: size }, (_, k) => make(start + k)), { ordered: false });
  }
};

const t0 = Date.now();
await insertAll('SubjectModel', SUBJECTS, (i) => ({
  _id: subjectIds[i], createdAt: someDate(i), updatedAt: someDate(i),
  dateOfBirth: new Date(1980, i % 12, 1), groupIds: [pick(groupIds, i)], sex: i % 2 ? 'MALE' : 'FEMALE'
}));
const sessionIds: ObjectId[] = [];
await insertAll('SessionModel', SESSIONS, (i) => {
  const _id = new ObjectId(); sessionIds.push(_id);
  return { _id, createdAt: someDate(i), updatedAt: someDate(i), date: someDate(i),
    groupId: pick(groupIds, i), subjectId: pick(subjectIds, i), userId: pick(userIds, i), type: 'IN_PERSON' };
});
const recordIds: ObjectId[] = [];
await insertAll('InstrumentRecordModel', RECORDS, (i) => {
  const _id = new ObjectId(); if (recordIds.length < FILES) recordIds.push(_id);
  return { _id, createdAt: someDate(i), updatedAt: someDate(i), date: someDate(i),
    data: { q1: i % 7, q2: `answer-${i % 50}`, q3: i % 2 === 0 }, computedMeasures: { score: i % 100 },
    groupId: pick(groupIds, i), subjectId: pick(subjectIds, i),
    instrumentId: i % 5 < 2 ? 'CORE_INTAKE' : pick(instrumentIds, i), sessionId: pick(sessionIds, i), pending: false };
});
await insertAll('AuditLogModel', AUDIT_LOGS, (i) => ({
  _id: new ObjectId(), action: pick(['CREATE', 'UPDATE', 'DELETE', 'LOGIN'], i), entity: pick(['SUBJECT', 'SESSION', 'INSTRUMENT_RECORD', 'USER'], i),
  groupId: pick(groupIds, i), timestamp: now - i * 1000, userId: pick(userIds, i)
}));
await insertAll('InstrumentRecordFileModel', FILES, (i) => ({
  _id: new ObjectId(), basename: `scan-${i % 10}`, createdAt: someDate(i), groupId: pick(groupIds, i),
  index: i % 4, name: `file-${i}.nii`, recordId: pick(recordIds, i), size: 1024 * (i % 500)
}));
console.log(`seeded ${(SUBJECTS + SESSIONS + RECORDS + AUDIT_LOGS + FILES).toLocaleString()} documents in ${((Date.now() - t0) / 1000).toFixed(1)}s`);

// Prisma must be able to read the documents back, or the shapes are wrong and the numbers are fiction.
const sanity = await prisma.instrumentRecord.count({ where: { subjectId: subjectIds[0] } });
if (sanity === 0) throw new Error('Prisma read back 0 records — document shape does not match the schema');
console.log(`prisma sanity check: ${sanity} records for subject 0\n`);

// ---------------------------------------------------------------- the queries services issue
const groupScope = { $in: groupIds };                    // accessibleQuery for a GROUP_MANAGER
const subject = subjectIds[0]!;
const minDate = new Date(now - 365 * 86_400_000);

const QUERIES = [
  { name: 'records for a subject (datahub)', collection: 'InstrumentRecordModel',
    filter: { $and: [{ date: { $gte: minDate } }, { groupId: groupScope }, { subjectId: subject }] },
    prisma: () => prisma.instrumentRecord.findMany({ where: { AND: [{ date: { gte: minDate } }, { groupId: { in: groupIds.map(String) } }, { subjectId: subject }] } }) },
  { name: 'records for a subject + instrument', collection: 'InstrumentRecordModel',
    filter: { $and: [{ subjectId: subject }, { instrumentId: 'CORE_INTAKE' }, { groupId: groupScope }] },
    prisma: () => prisma.instrumentRecord.findMany({ where: { AND: [{ subjectId: subject }, { instrumentId: 'CORE_INTAKE' }, { groupId: { in: groupIds.map(String) } }] } }) },
  { name: 'records by instrument (instruments/info, #1416)', collection: 'InstrumentRecordModel',
    filter: { instrumentId: 'CORE_INTAKE' },
    prisma: () => prisma.instrumentRecord.findMany({ where: { instrumentId: 'CORE_INTAKE' }, select: { subjectId: true }, distinct: ['subjectId'] }) },
  { name: 'record count for a group + date (summary)', collection: 'InstrumentRecordModel',
    filter: { groupId: groupIds[0], date: { $lte: new Date(now) } },
    prisma: () => prisma.instrumentRecord.count({ where: { groupId: String(groupIds[0]), date: { lte: new Date(now) } } }) },
  { name: 'records for a session', collection: 'InstrumentRecordModel',
    filter: { sessionId: sessionIds[0] },
    prisma: () => prisma.instrumentRecord.findMany({ where: { sessionId: String(sessionIds[0]) } }) },
  { name: 'sessions for a group', collection: 'SessionModel',
    filter: { groupId: groupIds[0] },
    prisma: () => prisma.session.findMany({ where: { groupId: String(groupIds[0]) } }) },
  { name: 'audit logs, group + sorted page', collection: 'AuditLogModel',
    filter: { groupId: groupIds[0] }, sort: { timestamp: -1 as const }, limit: 25,
    prisma: () => prisma.auditLog.findMany({ where: { groupId: String(groupIds[0]) }, orderBy: { timestamp: 'desc' }, skip: 0, take: 25 }) },
  { name: 'subjects for a group', collection: 'SubjectModel',
    filter: { groupIds: { $in: [groupIds[0]] } },
    prisma: () => prisma.subject.findMany({ where: { groupIds: { hasSome: [String(groupIds[0])] } } }) },
  { name: 'files for a record', collection: 'InstrumentRecordFileModel',
    filter: { recordId: recordIds[0] },
    prisma: () => prisma.instrumentRecordFile.findMany({ where: { recordId: String(recordIds[0]) } }) }
];

const median = (xs: number[]) => xs.slice().sort((a, b) => a - b)[Math.floor(xs.length / 2)]!;

async function measure() {
  const rows = [];
  for (const q of QUERIES) {
    let cursor = db.collection(q.collection).find(q.filter as any);
    if ((q as any).sort) cursor = cursor.sort((q as any).sort);
    if ((q as any).limit) cursor = cursor.limit((q as any).limit);
    const plan = await cursor.explain('executionStats');
    const stats = plan.executionStats;
    const stages: string[] = [];
    (function walk(node: any) {
      if (!node || typeof node !== 'object') return;
      if (typeof node.stage === 'string') stages.push(node.stage);
      for (const v of Object.values(node)) walk(v);
    })(plan.queryPlanner.winningPlan);
    const stage = stages.includes('IXSCAN') ? 'IXSCAN' : 'COLLSCAN';
    const sorted = stages.includes('SORT') ? ' + in-memory SORT' : '';

    await q.prisma();                                            // warm
    const times: number[] = [];
    for (let i = 0; i < REPEATS; i++) { const s = performance.now(); await q.prisma(); times.push(performance.now() - s); }
    rows.push({ name: q.name, examined: stats.totalDocsExamined, returned: stats.nReturned,
                ms: median(times), stage: stage + sorted });
  }
  return rows;
}

const statsFor = async () => {
  // A WiredTiger index is only sized accurately once its pages are on disk.
  await db.command({ fsync: 1 }).catch(() => {});
  const out: Record<string, { data: number; index: number; sizes: Record<string, number> }> = {};
  for (const c of ['InstrumentRecordModel', 'SessionModel', 'SubjectModel', 'AuditLogModel', 'InstrumentRecordFileModel']) {
    const s: any = await db.command({ collStats: c });
    out[c] = { data: s.size, index: s.totalIndexSize, sizes: s.indexSizes ?? {} };
  }
  return out;
};

console.log('measuring BEFORE (only _id_ indexes, as deployed today)…');
const before = await measure();
const statsBefore = await statsFor();

const buildStart = performance.now();
await ensureDatabaseIndexes(prisma as any);
const buildMs = performance.now() - buildStart;
const restartStart = performance.now();
await ensureDatabaseIndexes(prisma as any);
const restartMs = performance.now() - restartStart;

console.log('measuring AFTER…\n');
const after = await measure();
const statsAfter = await statsFor();

// ---------------------------------------------------------------- report
const fmtMs = (n: number) => (n < 1 ? '<1' : Math.round(n).toLocaleString());
console.log('| Query | docs examined → returned | | median ms | | plan |');
console.log('|---|---|---|---|---|---|');
console.log('| | before | after | before | after | before → after |');
before.forEach((b, i) => {
  const a = after[i]!;
  console.log(`| ${b.name} | ${b.examined.toLocaleString()} → ${b.returned.toLocaleString()} | ${a.examined.toLocaleString()} → ${a.returned.toLocaleString()} | ${fmtMs(b.ms)} | ${fmtMs(a.ms)} | ${b.stage} → ${a.stage} |`);
});

const mib = (n: number) => (n / 1024 / 1024).toFixed(2);
console.log('\n| Collection | data | index before (_id_ only) | index after | added | added / data |');
console.log('|---|---|---|---|---|---|');
let totalData = 0, totalAdded = 0;
for (const c of Object.keys(statsBefore)) {
  const d = statsAfter[c]!.data, ib = statsBefore[c]!.index, ia = statsAfter[c]!.index;
  totalData += d; totalAdded += ia - ib;
  console.log(`| ${c} | ${mib(d)} MiB | ${mib(ib)} MiB | ${mib(ia)} MiB | ${mib(ia - ib)} MiB | ${((ia - ib) / d * 100).toFixed(1)}% |`);
}
console.log(`| **total** | **${mib(totalData)} MiB** | | | **${mib(totalAdded)} MiB** | **${(totalAdded / totalData * 100).toFixed(1)}%** |`);
console.log('\nper-index sizes after:');
for (const [c, v] of Object.entries(statsAfter)) {
  for (const [name, size] of Object.entries(v.sizes)) console.log(`  ${name}: ${(size / 1024).toFixed(0)} KiB`);
}
console.log(`\nindex build on this dataset: ${(buildMs / 1000).toFixed(2)}s`);
console.log(`subsequent restart (all present): ${Math.round(restartMs)}ms`);

await prisma.$disconnect(); await mc.close(); await rs.stop();
```

</details>

## Verification

`ensureDatabaseIndexes` run against a real MongoDB replica set:

| | |
|---|---|
| fresh database | 17 indexes created; `assignmentId` is `unique sparse` |
| restart | 17, no rebuilds (idempotent) |
| three records with no assignment | all CREATED (was `P2002` on #2) |
| duplicate group name | REJECTED `P2002` — the constraint is enforced for the first time |
| after a `db push` non-sparse index | replaced with the sparse one; inserts succeed again |

Suites, from the repo root:

- `pnpm lint` — 33/33 green.
- `pnpm test` — 741 passed. 3 failures in `apps/web/src/__tests__/start-session-form.test.tsx`.
- `pnpm test:e2e` — 142/148 passed, including the new spec and both firefox `@smoke` runs. 6 failures across `start-session.spec.ts` and `datahub.spec.ts`.

**The 9 failures above are pre-existing on this branch's parent and unrelated to this change** — all of them are in the custom-identifier combobox area from #1508, and they fail identically with this branch's changes stashed.

## Tests

- `apps/api/src/core/__tests__/prisma.spec.ts` — 13 unit tests covering name derivation, the sparse `assignmentId` index, compound-field ordering, the non-sparse repair path, a missing collection, and loud failure. Two of them are the regression guard #1406 asks for: they parse `schema.prisma` and assert it and `DATABASE_INDEXES` agree in both directions, so neither list can drift from the other.
- `testing/src/specs/instrument-completion.spec.ts` — administers one instrument twice for a subject and asserts both records land, which is the end-to-end form of the `P2002` regression.

`apps/api/AGENTS.md` is updated in the same commit: its Prisma section previously said `prisma db push` was the way to sync the schema.

https://claude.ai/code/session_01RWzUZAuSJMy2CVjL2kDSgA

### DIFF
--- a/apps/api/AGENTS.md
+++ b/apps/api/AGENTS.md
@@ -122,8 +122,26 @@ or `ValidObjectIdPipe` explicitly — see `src/audit/audit.controller.ts` and
 
 ## Prisma / database
 
-Provider is **MongoDB, so there are no migrations** — `prisma db push` only, and nothing to check in
-after a schema change beyond `schema.prisma` itself.
+Provider is **MongoDB, so there are no migrations**. Nothing is checked in after a schema change
+beyond `schema.prisma` itself, and **nothing in this app runs `prisma db push`**. This workspace
+defines no `db:push` script, so the `db:push` that `dev`, `dev:test`, `lint` and `test:e2e` depend on
+resolves to nothing here; the only workspace that owns one is `apps/gateway`, whose datasource is
+SQLite. The Dockerfile only builds.
+
+**Indexes are created at startup instead**, by `ensureDatabaseIndexes` in `src/core/prisma.ts`,
+which the client factory awaits after `$connect()`. MongoDB creates a collection on first insert
+with only its `_id_` index, so without that step a deployed instance scans whole collections and
+enforces none of the schema's unique constraints. Adding an index means editing **both**
+`schema.prisma` and `DATABASE_INDEXES`, which
+`src/core/__tests__/prisma.spec.ts` asserts agree — `db push` drops any index the schema
+does not declare, and a runtime-only index would not survive one.
+
+**Do not point `prisma db push` at a database this application uses.** `InstrumentRecord.assignmentId`
+is optional but must carry `@unique`, because Prisma requires it on the defining side of the 1:1
+relation to `Assignment`. `db push` builds that index non-sparse, MongoDB indexes a missing field as
+null, and the second record collected outside a remote assignment is then rejected with a duplicate
+key error. `ensureDatabaseIndexes` creates it sparse and replaces a non-sparse one it finds, so an
+instance recovers on its next boot — but the push is still what broke it.
 
 `datasource db { url = env("_") }` in `apps/api/prisma/schema.prisma` is **deliberate**; the real
 connection string is built at runtime in `src/core/prisma.ts`. Do not "fix" it.
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -58,6 +58,9 @@ model AuditLog {
   user      User?          @relation(fields: [userId], references: [id])
   userId    String?        @db.ObjectId
 
+  @@index([timestamp])
+  @@index([groupId, timestamp])
+  @@index([userId, timestamp])
   @@map("AuditLogModel")
 }
 
@@ -78,6 +81,8 @@ model Assignment {
   url               String
   encryptionKeyPair EncryptionKeyPair
 
+  @@index([groupId])
+  @@index([subjectId])
   @@map("AssignmentModel")
 }
 
@@ -92,7 +97,6 @@ enum SubjectIdentificationMethod {
   CUSTOM_ID
   PERSONAL_INFO
 }
-
 
 type GroupSettings {
   defaultIdentificationMethod   SubjectIdentificationMethod
@@ -151,30 +155,37 @@ model Group {
 /// Instrument Records
 
 model InstrumentRecord {
-  createdAt        DateTime               @default(now()) @db.Date
-  updatedAt        DateTime               @updatedAt @db.Date
-  id               String                 @id @default(auto()) @map("_id") @db.ObjectId
+  createdAt          DateTime               @default(now()) @db.Date
+  updatedAt          DateTime               @updatedAt @db.Date
+  id                 String                 @id @default(auto()) @map("_id") @db.ObjectId
   /// [ComputedMeasures]
-  computedMeasures Json?
-  data             Json?
-  date             DateTime               @db.Date
-  files            InstrumentRecordFile[]
-  group            Group?                 @relation(fields: [groupId], references: [id])
-  groupId          String?                @db.ObjectId
-  subject          Subject                @relation(fields: [subjectId], references: [id])
-  subjectId        String
-  instrument       Instrument             @relation("InstrumentRecords", fields: [instrumentId], references: [id])
-  instrumentId     String
+  computedMeasures   Json?
+  data               Json?
+  date               DateTime               @db.Date
+  files              InstrumentRecordFile[]
+  group              Group?                 @relation(fields: [groupId], references: [id])
+  groupId            String?                @db.ObjectId
+  subject            Subject                @relation(fields: [subjectId], references: [id])
+  subjectId          String
+  instrument         Instrument             @relation("InstrumentRecords", fields: [instrumentId], references: [id])
+  instrumentId       String
   // The series instrument, if any, that orchestrated this record's collection. Null for records
   // collected outside a series. The series id encodes the exact ordered composition of its forms.
-  seriesInstrument   Instrument?          @relation("SeriesInstrumentRecords", fields: [seriesInstrumentId], references: [id])
+  seriesInstrument   Instrument?            @relation("SeriesInstrumentRecords", fields: [seriesInstrumentId], references: [id])
   seriesInstrumentId String?
-  assignment       Assignment?            @relation(fields: [assignmentId], references: [id])
-  assignmentId     String?                @unique
-  session          Session                @relation(fields: [sessionId], references: [id])
-  sessionId        String                 @db.ObjectId
-  pending          Boolean?
+  assignment         Assignment?            @relation(fields: [assignmentId], references: [id])
+  // Prisma requires @unique here for the 1:1 relation, but builds the index non-sparse, and MongoDB
+  // indexes a missing field as null — so a second record without an assignment would collide.
+  // ensureDatabaseIndexes in src/core/prisma.ts creates this one sparse instead.
+  assignmentId       String?                @unique
+  session            Session                @relation(fields: [sessionId], references: [id])
+  sessionId          String                 @db.ObjectId
+  pending            Boolean?
 
+  @@index([subjectId, instrumentId, date])
+  @@index([groupId, date])
+  @@index([instrumentId])
+  @@index([sessionId])
   @@map("InstrumentRecordModel")
 }
 
@@ -190,6 +201,7 @@ model InstrumentRecordFile {
   recordId  String           @db.ObjectId
   size      Int
 
+  @@index([recordId])
   @@map("InstrumentRecordFileModel")
 }
 
@@ -233,19 +245,19 @@ model Instrument {
 // Instrument Repos
 
 model InstrumentRepo {
-  createdAt     DateTime   @default(now()) @db.Date
-  updatedAt     DateTime   @updatedAt @db.Date
-  id            String     @id @default(auto()) @map("_id") @db.ObjectId
+  createdAt     DateTime  @default(now()) @db.Date
+  updatedAt     DateTime  @updatedAt @db.Date
+  id            String    @id @default(auto()) @map("_id") @db.ObjectId
   // Display label only (derived from the repo name); not unique because two owners may have repos with
   // the same name (e.g. alice/forms and bob/forms). `url` is the true identity and is unique.
   name          String
   owner         String
   repoName      String
-  url           String     @unique
+  url           String    @unique
   instrumentIds String[]
-  groupIds      String[]   @db.ObjectId
-  groups        Group[]    @relation(fields: [groupIds], references: [id])
-  lastSyncedAt  DateTime?  @db.Date
+  groupIds      String[]  @db.ObjectId
+  groups        Group[]   @relation(fields: [groupIds], references: [id])
+  lastSyncedAt  DateTime? @db.Date
   // AES-256-GCM encrypted GitHub personal access token, used to access private repositories.
   accessToken   String?
 
@@ -273,6 +285,7 @@ model Subject {
   instrumentRecords InstrumentRecord[]
   assignments       Assignment[]
 
+  @@index([groupIds])
   @@map("SubjectModel")
 }
 
@@ -327,6 +340,7 @@ model User {
   email                 String?
   disabled              Boolean?
 
+  @@unique([username])
   @@map("UserModel")
 }
 
@@ -350,6 +364,8 @@ model Session {
   userId            String?            @db.ObjectId
   type              SessionType
 
+  @@index([groupId, date])
+  @@index([subjectId])
   @@map("SessionModel")
 }
 
--- a/apps/api/src/core/__tests__/prisma.spec.ts
+++ b/apps/api/src/core/__tests__/prisma.spec.ts
@@ -1,0 +1,181 @@
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+import { describe, expect, it, vi } from 'vitest';
+
+import { buildIndexName, DATABASE_INDEXES, ensureDatabaseIndexes } from '../prisma';
+
+type IndexSpecification = {
+  key: { [field: string]: number };
+  name: string;
+  sparse?: boolean;
+  unique?: boolean;
+};
+
+/**
+ * A client whose listIndexes reports `existing`, recording every command it is asked to run. A
+ * collection named in `rejectCreateOn` fails its createIndexes, as a duplicate would make it.
+ */
+function mockClient(existing: { [collection: string]: IndexSpecification[] } = {}, rejectCreateOn?: string) {
+  const commands: { [key: string]: any }[] = [];
+  const $runCommandRaw = vi.fn((command: { [key: string]: any }) => {
+    commands.push(command);
+    if (typeof command.listIndexes === 'string') {
+      const indexes = existing[command.listIndexes];
+      if (!indexes) {
+        throw Object.assign(new Error('ns does not exist'), { code: 26 });
+      }
+      return Promise.resolve({ cursor: { firstBatch: [{ key: orderedKey('_id'), name: '_id_' }, ...indexes] } });
+    }
+    if (rejectCreateOn && command.createIndexes === rejectCreateOn) {
+      throw new Error('E11000 duplicate key error');
+    }
+    return Promise.resolve({ ok: 1 });
+  });
+  return { client: { $runCommandRaw }, commands };
+}
+
+/** Built from a list because `perfectionist/sort-objects` alphabetizes an object literal, and the
+ * order of an index's fields is the thing under test. */
+const orderedKey = (...fields: string[]): { [field: string]: number } =>
+  Object.fromEntries(fields.map((field) => [field, 1]));
+
+const createdIndexes = (commands: { [key: string]: any }[], collection: string): IndexSpecification[] =>
+  commands.find((command) => command.createIndexes === collection)?.indexes ?? [];
+
+const findIndex = (commands: { [key: string]: any }[], collection: string, name: string) =>
+  createdIndexes(commands, collection).find((index) => index.name === name);
+
+describe('buildIndexName', () => {
+  // db push drops an index the schema does not name, so a name that differs from Prisma's leaves
+  // two copies of the same index and the application's one is deleted on the next push.
+  it('should reproduce the suffix prisma db push gives a non-unique index', () => {
+    expect(buildIndexName({ collection: 'SessionModel', fields: ['groupId', 'date'] })).toBe(
+      'SessionModel_groupId_date_idx'
+    );
+  });
+
+  it('should reproduce the suffix prisma db push gives a unique index', () => {
+    expect(buildIndexName({ collection: 'UserModel', fields: ['username'], unique: true })).toBe(
+      'UserModel_username_key'
+    );
+  });
+});
+
+describe('ensureDatabaseIndexes', () => {
+  it('should create every declared index', async () => {
+    const { client, commands } = mockClient();
+    await ensureDatabaseIndexes(client);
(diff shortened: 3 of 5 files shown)


gdevenyi:
@joshunrau benchmarks?

joshunrau:
@gdevenyi benchmarks added

PR #1520, opened 2026-09-01
feat(playground): show the ODC version in the actions menu

thomasbeaudry:
Adds the ODC version to the bottom of the playground's `...` menu, under a separator.

```
GitHub / Login / Upload Bundle / User Settings / Storage Usage
Delete Instrument / Restore Defaults
─────────────────
ODC Version 2.2.3
```

## How the version gets there

Read from the root `package.json` in `vite.config.ts` and supplied as `__APP_VERSION__`, alongside the existing `__GITHUB_REPO_URL__` define.

Not the `__RELEASE__` define `apps/web` uses: `@opendatacapture/release-info` pulls in `@opendatacapture/schemas`, which ships raw TypeScript, so importing it here fails to load the config unless the playground's `dev` and `build` scripts run Vite under `NODE_OPTIONS='--import=tsx'` the way `apps/web` does. Reading the file directly avoids changing how this app builds and adds no dependency.

## Tests

None. The playground has no `vitest.config.ts`, and `testing/`'s Playwright config only starts api, gateway and web — so there is no harness this could hook into without adding a webServer entry for an app that downloads an 11 MB wasm on boot.

Verified by hand instead: loaded the dev server, opened the menu, confirmed `ODC Version 2.2.3` renders with no page errors.

### DIFF
--- a/apps/playground/src/components/Header/ActionsDropdown/ActionsDropdown.tsx
+++ b/apps/playground/src/components/Header/ActionsDropdown/ActionsDropdown.tsx
@@ -81,6 +81,10 @@ export const ActionsDropdown = () => {
               {t({ en: 'Restore Defaults', fr: 'Restaurer les paramètres par défaut' })}
             </button>
           </DropdownMenu.Item>
+          <DropdownMenu.Separator />
+          <DropdownMenu.Label className="text-muted-foreground text-xs font-normal">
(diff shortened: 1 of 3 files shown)


thomasbeaudry:
image

joshunrau (approved):

Issue #1500, opened 2026-08-06
feat: add group-scoped bulk remote assignments


thomasbeaudry:
## Description

Clinicians currently create remote assignments one subject at a time. Add a group-scoped wizard that accepts up to 500 subject identifiers, previews and validates the data, selects one accessible instrument and expiry date, reports conflicts, and creates the assignments in one operation.

This is the first release of bulk remote assignments. It deliberately supports **existing subjects in the selected group only**. Creating/linking subjects and sending assignment emails in bulk need separate designs because they change authorization, ownership, delivery, and throttling semantics.

Estimated difficulty: **High**.

## User flow

Add `/group/bulk-remote-assignments` with six explicit states:

1. **Source** — upload `.csv`/`.xlsx`, or paste comma-, tab-, or semicolon-separated data.
2. **Map and validate** — preview the first four rows, correct header mappings, and report row-numbered validation errors.
3. **Instrument** — choose one instrument from the current group's accessible instruments using `InstrumentShowcase`.
4. **Expiry** — choose a future expiry, defaulted with `getDefaultAssignmentExpiry()`.
5. **Review** — show each subject as `READY`, `CONFLICT`, or `SUBJECT_UNAVAILABLE`; conflicts default to skip and may be explicitly allowed as duplicates.
6. **Results** — show `CREATED`, `SKIPPED`, or `FAILED` per source row; allow formula-safe CSV download and starting over.

`SUBJECT_UNAVAILABLE` intentionally does not distinguish a missing subject from one outside the selected group. This avoids leaking subject existence across groups.

## Input and privacy rules

Support two modes:

- **ID mode:** one mapped `subjectId` column.
- **PII mode:** mapped `firstName`, `lastName`, `dateOfBirth`, and `sex` columns. Resolve IDs in the browser with `generateSubjectHash()` from `@opendatacapture/subject-utils`.

Header matching is case-insensitive and normalizes whitespace, punctuation, and diacritics. Recognize English/French aliases for the subject ID and four PII fields. Each canonical field may be mapped at most once; extra columns, including email, are ignored in v1.

Validation must:

- reject empty/invalid parser output and surface Papa Parse/XLSX errors;
- skip fully empty rows;
- reject ambiguous PII dates; CSV/paste dates must be ISO `YYYY-MM-DD`;
- normalize sex only through the existing subject contract and reject unknown values;
- reject duplicate resolved subject IDs;
- reject more than 500 resolved subjects;
[...]

PR #1528, opened 2026-09-10
feat(web): add the bulk remote assignment wizard

thomasbeaudry:
The web half of group-scoped bulk remote assignments (#1500), on top of the API and gateway merged in #1521.

`/group/bulk-remote-assignments` walks three steps: choose subjects, choose instruments and expiries, review. The batch is all-or-nothing, so review is the last point at which anything can be changed.

## Choosing subjects

One source at a time: pick from the group's subjects, upload a `.csv`/`.tsv`/`.xlsx`, or paste delimited data. Headers are matched case- and accent-insensitively against English and French aliases, and the mapping is shown back before anything is sent.

In PII mode the identifiers are derived in the browser with `generateSubjectHash`, so only derived ids ever reach the API. `xlsx` is imported dynamically, so only a user who picks a workbook pays for it.

## Results

Each subject's assignment link, with a per-row copy, copy-all, and a CSV. When the batch came from a file, the export echoes back the row the user supplied beside the link, since a derived identifier is a hash and cannot otherwise be matched to a person. The clipboard carries the same rows as the file, from the same builder, so the two cannot drift.

## Admin toggle

`isBulkRemoteAssignmentsEnabled` sits beside the uploader flag. With it off the group links stay exactly where they are today; with it on they collapse into a `Group Actions` menu with the new page. Children are gated independently, so losing one ability does not hide the others. The route guards itself on both the toggle and the gateway rather than only being hidden from the nav.

## Worth a reviewer's attention

- **`resolveSubjectIds` re-scopes a pasted identifier.** A custom id is stored as `Group$id` but displayed and exported without the prefix, so a user pasting what they see would otherwise match nothing. Digests and already-scoped values are left alone.
- **The exported identifier is deliberately not truncated.** The app displays a short form everywhere, but a prefix resolves to no record, so the file carries the full value.
- **No component tests for the wizard.** The parsing utility and the hooks are covered; the steps themselves are not.

## Follow-up found while testing

`StartSessionForm`, `IdentificationForm` and `utils/upload.ts` all scope an entered identifier unconditionally. Pasting a digest into Start Session produces `Group$<digest>`, which matches nothing, and the upload path creates any subject it cannot find — so this silently makes a duplicate. Pre-existing on `main` and untouched here; the guard added in this PR is what those three call sites need, lifted into `subject-utils`.


apps/web/src/hooks/useNavItems.ts
```diff
@@ -81,21 +81,52 @@ export function useNavItems() {
...
+      isBulkEnabled &&
+      currentGroup &&
+      config.setup.isGatewayEnabled &&
+      ability?.can('create', 'Assignment') &&
+      ability.can('read', 'Assignment') &&
+      ability.can('read', 'Subject')
+    ) {
+      groupItems.push({
+        icon: SendIcon,
+        label: t({ en: 'Bulk Remote Assignments', fr: 'Tâches à distance en lot' }),
```
joshunrau: I would say we should just name this "Remote Assignments", it is more aesthetic

apps/api/src/setup/setup.service.ts
```diff
@@ -61,6 +61,7 @@ export class SetupService {
       activeLanguages: fallbackLanguage ? [fallbackLanguage, ...otherLanguages] : DEFAULT_ACTIVE_LANGUAGES,
       branding: branding.success ? branding.data : null,
       defaultAssignmentDurationDays: savedOptions?.defaultAssignmentDurationDays ?? null,
+      isBulkRemoteAssignmentsEnabled: savedOptions?.isBulkRemoteAssignmentsEnabled !== false,
```
joshunrau: just verify that is it impossible for `isBulkRemoteAssignmentsEnabled` to be set to true, if `isGatewayEnabled` is not true

apps/web/src/components/BulkRemoteAssignmentWizard/BulkRemoteAssignmentWizard.tsx
```diff
@@ -0,0 +1,327 @@
...
+    });
+
+  const copyLinks = async () => {
+    try {
+      if (!navigator.clipboard) {
+        throw new Error('Clipboard unavailable');
+      }
+      await navigator.clipboard.writeText(toResultTsv(resultRows()));
+      setDidCopy(true);
+    } catch {
```
joshunrau: remove the try catch, we can just add notification and return if clipboard is not available for some reason

apps/web/src/components/BulkRemoteAssignmentWizard/BulkRemoteAssignmentWizard.tsx
```diff
@@ -0,0 +1,327 @@
...
+  groupId,
+  groupName,
+  instruments,
+  subjectIdDisplayLength,
+  subjects
+}: BulkRemoteAssignmentWizardProps) => {
+  const { t } = useTranslation();
+  // Each piece of the batch is held here rather than in the step that edits it, so moving between
+  // steps - by Back or by breadcrumb - never discards work the user has already done.
+  const [step, setStep] = useState<WizardStep>('SOURCE');
```
joshunrau: extract all of this to a single hook

joshunrau: only make public what needs to be public

apps/web/src/components/BulkRemoteAssignmentWizard/BulkRemoteAssignmentWizard.tsx
```diff
@@ -0,0 +1,283 @@
...
+        onSuccess: (created) => {
+          setAssignments(created);
+          setStep('DONE');
+        }
+      }
+    );
+  };
+
+  return (
+    <div className="mx-auto flex max-w-[70rem] flex-col gap-6" data-testid="bulk-remote-assignment-wizard">
```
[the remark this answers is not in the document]
joshunrau: yes, test the hook I requested

apps/web/src/components/BulkRemoteAssignmentWizard/TimepointsStep.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const [instrumentId, setInstrumentId] = useState('');
+  const [expiresAt, setExpiresAt] = useState(defaultExpiresAt);
+
+  // An instrument already in the list cannot be added again: two live assignments for the same
+  // subject and instrument is exactly the conflict the API refuses.
+  const available = instruments.filter(
+    (instrument) => !timepoints.some((timepoint) => timepoint.instrumentId === instrument.id)
+  );
+
+  const todayISO = new Date().toISOString().split('T')[0]!;
```
joshunrau: use toBasicIsoString from libjs

apps/web/src/components/BulkRemoteAssignmentWizard/TimepointsStep.tsx
```diff
@@ -0,0 +1,159 @@
...
+              {t({ en: 'Expires on', fr: 'Expire le' })}
+            </label>
+            <Input
+              className="w-full"
+              data-testid="bulk-expiry-input"
+              id="bulk-expiry"
+              type="date"
+              value={expiresAt}
+              onChange={(event) => setExpiresAt(event.target.value)}
+            />
```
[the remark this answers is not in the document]
joshunrau: this is wrong, ignore

joshunrau (changes requested):
overall comments
- should use title case when applicable consistent with the rest of the app
- the header thing (1. Subjects, 2. Instruments, etc.) looks bad, especially on large screens. Open it in the browser and improve the UI
- need padding for table - it needs to be opened in the browser and examined for UI
- need shadow for table or something, it looks very bad now
-

apps/web/src/hooks/useNavItems.ts
thomasbeaudry: Renamed to "Remote Assignments" in 0c41a5401 — nav item, page heading and the admin setting ("Enable Remote Assignments"), so the three agree.

apps/api/src/setup/setup.service.ts
thomasbeaudry: Enforced in 0c41a5401. `updateState` now throws `ForbiddenException` if `isBulkRemoteAssignmentsEnabled` is set true while `GATEWAY_ENABLED` is false, and `getState` derives it from the gateway, so a document written while a gateway existed still reports false without one. Four cases added in `setup.service.spec.ts`, including disabling it without a gateway (still allowed, so an instance is never stuck with it on).

apps/web/src/components/BulkRemoteAssignmentWizard/BulkRemoteAssignmentWizard.tsx
thomasbeaudry: Done in 0c41a5401 — `copyLinks` notifies and returns when `navigator.clipboard` is absent, no try/catch.

apps/web/src/components/BulkRemoteAssignmentWizard/BulkRemoteAssignmentWizard.tsx
thomasbeaudry: Extracted to `apps/web/src/hooks/useBulkAssignmentWizard.ts` in 0c41a5401. The component now only renders: every step value, the mutations and the step transitions live in the hook. Raw uploaded rows stay private to it — `describeSubject` and `resultRows` are what the component gets instead.

apps/web/src/components/BulkRemoteAssignmentWizard/BulkRemoteAssignmentWizard.tsx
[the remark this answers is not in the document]
thomasbeaudry: Added `apps/web/src/hooks/__tests__/useBulkAssignmentWizard.test.ts` in 0c41a5401: 12 cases covering source switching, back/breadcrumb state retention, the preflight refusal landing on review, transport failure vs refusal, the expiry payload, results, reset, and both clipboard paths.

apps/web/src/components/BulkRemoteAssignmentWizard/TimepointsStep.tsx
thomasbeaudry: Done in 0c41a5401 — the export is `toBasicISOString`, now used in `TimepointsStep` and in the route that computes the default expiry.

thomasbeaudry:
@joshunrau addressed in 0c41a5401 (rebased onto 120fdf22c).

**Inline comments** — replied on each thread: nav renamed to "Remote Assignments", wizard state extracted into `useBulkAssignmentWizard` with 12 unit tests, clipboard try/catch removed, `toBasicISOString` used for both date spots, and the toggle now refuses to be enabled without the gateway (`ForbiddenException` in `updateState`, plus `getState` deriving it from the gateway so a document written under one cannot report it on afterwards). I skipped the TimepointsStep date-constraint thread you marked "ignore".

**Overall comments**
- Title case applied across the wizard: step titles, tab labels, buttons, table headers and field labels. French left in sentence case, as elsewhere.
- Step header rebuilt: the "1. Subjects / 2. Instruments" buttons are now numbered markers joined by a rule, capped at `max-w-md` so it no longer stretches across a wide card.
- Tables: subject picker, column mapping, preview and results all get a border, `shadow-sm` and taller row padding.

**Checks:** web + api lint clean; 331/331 web unit tests; 325/325 api tests.

Issue #1527, opened 2026-09-09
Manage user page does not display users permissions upon initial load.


david-roper:
### What Happened?

Upon the admins initial load of a user's manage user form their permissions are not displayed, and instead a blank set of permissions is presented. This is similar to the subject table loading issue we had previously. #1330. The issue solves itself if the admin reloads the manage user form. 

### What Did You Expect?

I expect the a users permission to be correctly displayed up this intial loading of the manage user form. 

### Operating System

Ubuntu 22.04

### Browser (if applicable)

Firefox

### Steps to Reproduce

Login as admin
navigate to admin user page
select a user with permissions

joshunrau:
Closing as done: PR #1531 fixed the root cause — the permissions fieldset was recreated as a new literal on every background refetch, wiping loaded state. Now memoized.

PR #1531, opened 2026-09-10
Load permission properly on initial manageUserForm selection

david-roper:
## Show a user's existing permissions when the manage-user sheet is opened

Opening **Manage Users → a user** showed an empty permission row even when that user held
permissions. Reopening the sheet later showed them correctly, which made it look like a
loading problem.

It was not. The permissions loaded fine — they were erased immediately afterwards.

### Root cause

libui's `RecordArrayField` resets itself whenever its `fieldset` prop changes identity:

```tsx
const fieldsetRef = useRef(fieldset);
useEffect(() => {
  if (fieldsetRef.current !== fieldset) {
    setArrayValue([createNewRecord()]);   // ← discards the loaded permissions
    fieldsetRef.current = fieldset;
  }
}, [fieldset]);
```

UpdateUserForm declared that fieldset as an inline object literal in its JSX, so every
render handed the field a structurally identical object with a fresh identity, and the
field blanked itself. The form mounts with the correct values; the next render destroys them.

On a first visit that next render is guaranteed: queryClient sets no staleTime, so
useUsersQuery / useGroupsQuery background-refetch on mount. When the refetch lands,
groupsQuery.data is a new array, the effect rebuilds data, and UpdateUserForm re-renders.
On a later open the queries are already fresh, nothing refetches, and the values survive —
hence "it works the second time".

key={JSON.stringify(initialValues)} did not help: the serialized string is unchanged for the
same user, so the Form never remounted to re-seed itself.

###  Why this mattered beyond the empty row
The reset left the field holding [{ action: undefined, subject: undefined }], which the
schema's .transform collapses to []. An admin who opened the sheet and saved any change —
an email, a group — silently stripped that user's additional permissions, having never been
shown them. Data loss, not just a rendering glitch.

#### The fix
apps/web/src/routes/_app/admin/users/index.tsx — the permission fieldset is now useMemo'd
on [resolvedLanguage], the only input that can change it, giving it a stable identity across
re-renders.

Memoized at the fieldset level rather than around the whole content array on purpose:
groupOptions is rebuilt on every refetch, so a content-level memo keyed on it would keep
churning and re-introduce the bug.

#### Tests

Unit — apps/web/src/__tests__/record-array-initial-values.test.tsx pins both halves of the libui contract this fix depends on: a stable fieldset renders, keeps and submits its seeded records across a parent re-render; an unstable one discards them. Written against the libui contract rather than the route, following data-table-server-mode.test.tsx, because route files export only Route.

E2E — admin-management.spec.ts seeds a user holding read Subject, opens the sheet, asserts the selects show it, then forces a re-render through the UI (open the delete dialog, click No) and asserts they still do.

ApiClient.updateUser added to testing/src/support/api-client.ts: additionalPermissions is on the update schema and not the create one, so seeding such a user takes two calls.

The e2e test was confirmed to fail on the unfixed code with the exact reported symptom
(Expected "Read", Received "") and to pass with the fix.

#### Verification
pnpm lint — 33/33 tasks successful
pnpm test — 796 passed
pnpm test:e2e (chromium) — 146 passed

Closes issue #1527

### DIFF
--- a/apps/web/src/routes/_app/admin/users/index.tsx
+++ b/apps/web/src/routes/_app/admin/users/index.tsx
@@ -70,6 +70,91 @@ const UpdateUserForm: React.FC<{
   const [isConfirmDeleteOpen, setIsConfirmDeleteOpen] = useState(false);
   const { applyGeneratedPassword, generatedPassword, generatePassword, isGeneratedPassword } = usePasswordGenerator();
 
+  // libui's `record-array` field resets itself to a single blank record whenever its `fieldset`
+  // changes identity, so an inline literal would discard the permissions it was seeded with on the
+  // next render of this component -- which a background refetch of either query triggers.
+  const additionalPermissionsFieldset = useMemo(
+    () => ({
+      action: {
+        kind: 'string' as const,
+        label: t({
+          en: 'Action',
+          fr: 'Action'
+        }),
+        options: {
+          create: t({
+            en: 'Create',
+            fr: 'Créer'
+          }),
+          delete: t({
+            en: 'Delete',
+            fr: 'Supprimer'
+          }),
+          manage: t({
+            en: 'Manage (All)',
+            fr: 'Gérer (Tout)'
+          }),
+          read: t({
+            en: 'Read',
+            fr: 'Lire'
+          }),
+          update: t({
+            en: 'Update',
+            fr: 'Modifier'
+          })
+        },
+        variant: 'select' as const
+      },
+      subject: {
+        kind: 'string' as const,
+        label: t({
+          en: 'Resource',
+          fr: 'Ressource'
+        }),
+        options: {
+          all: t({
+            en: 'All',
+            fr: 'Tous'
+          }),
+          Assignment: t({
+            en: 'Assignment',
+            fr: 'Assignation'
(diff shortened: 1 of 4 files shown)


apps/web/src/routes/_app/admin/users/index.tsx
```diff
@@ -70,6 +70,91 @@ const UpdateUserForm: React.FC<{
   const [isConfirmDeleteOpen, setIsConfirmDeleteOpen] = useState(false);
   const { applyGeneratedPassword, generatedPassword, generatePassword, isGeneratedPassword } = usePasswordGenerator();
 
+  // libui's `record-array` field resets itself to a single blank record whenever its `fieldset`
+  // changes identity, so an inline literal would discard the permissions it was seeded with on the
+  // next render of this component -- which a background refetch of either query triggers.
+  const additionalPermissionsFieldset = useMemo(
```
joshunrau: add type

joshunrau: useMemo<T>

joshunrau (changes requested):

joshunrau (approved):

PR #1549, opened 2026-09-17
fix(web): correct autofill and password-field behaviour on the user edit sheet

thomasbeaudry:
## Summary

Three fixes to the manage-user sheet on `/admin/users`.

- **Saved credentials no longer autofill the form.** The sheet is shaped like a sign-in form, so the browser filled the signed-in administrator's own saved password into the password fields and their username into the email field — on a form that edits somebody else's account. libui's `Input` hardcodes `autocomplete="off"`, which browsers ignore for credential fields, and libui's field configuration exposes no autocomplete option, so `useSuppressPasswordAutofill` sets `autocomplete="new-password"` on the rendered password inputs. That is the value password managers honour as "not the saved credential", and it also stops the form being read as a sign-in at all, which is what leaves the email field alone.
- **Clearing the password no longer fails validation.** A blank field was read as an empty password rather than as no password, so it was rejected for being too short, and its confirmation was rejected for not matching. `$OptionalPassword` normalizes `''` to `undefined` before any other rule sees it.
- **The strength meter no longer flags an untouched field.** The password here is optional, so blank is its normal state, but libui scores an empty field zero and paints the meter red — which reads as a rejection of a field the administrator is deliberately leaving alone. The meter is dropped from this sheet; strength is still enforced by the schema on submit. The create-user form, where the password is required, keeps it.

The password labels are now "Set new password" and "Confirm new password", so the fields read as optional rather than as the account's current credentials.

## Test plan

- [ ] `pnpm lint`
- [ ] `pnpm test` — unit tests cover `useSuppressPasswordAutofill` and `$OptionalPassword`
- [ ] `pnpm test:e2e` — `admin-management.spec.ts` asserts both password inputs carry `autocomplete="new-password"`, and that a password typed and then cleared still saves
- [ ] Manually: open `/admin/users`, manage a user, and confirm no field arrives prefilled and that the password fields are unmarked while blank


joshunrau (changes requested):
I changed this completely when I added a new page to manage the user (with the new permissions editor interface). Could you please adjust this to implement your fix on the new page?

Issue #1355, opened 2026-05-07
Allow for group scoped permissions for standard and group manager users


david-roper:
### Current Implementation and Limitations

Currently the permissions a user has are not attached to any group. There are cases where we would like a user to have certain permissions for one group and different ones for another. It should also be mandatory for a non-admin user to have a group to make it so they don't have admin level views of the datahub, instruments, etc.

### Associated Application Components

Client, Server

### Proposed Solution

Edit the manage user form to have the group selection be mandatory for non-admin users. When a group is selected have a record array for the permission of that group appear. 

On server side adjust the ability factory to include a new condition, where they have to part of a specific group to have the permissions. 

Create a new type in the `prisma-json-types-generator.d.ts` which is the condition for the group and the users permissions for it.

Add this conditional type to the user schema as a json. 

create tests to make sure the user can have different permission for each group.

### Estimated Difficulty

High

### Priority

Medium

PR #1553, opened 2026-09-19
feat: group-scoped user permissions managed on a dedicated user page

joshunrau:
## Summary

Admins could only grant "additional permissions" instance-wide: `AuthRule` had nowhere to put a scope, the ability factory applied every grant unconditionally, and the sheet on `/admin/users` told admins to "use the API" for anything group-specific, which the API could not do either. Granting `read Subject` to a standard user of one group silently exposed every group's subjects, and a granted `update User` let that user `PATCH` themselves to `manage all`.

Every grant is now confined to one of the user's groups by default, an unscoped grant is a deliberate and visible choice, and the whole thing is managed on a dedicated page instead of the sheet.

## API

- `AuthRule.groupId` (`String?`): null means every group, which is what every row written before the field existed reads back as. No migration; an integration suite pins the legacy read-back.
- `AbilityFactory` turns a non-null `groupId` into that model's group condition through `GROUP_SCOPED_CONDITIONS`, typed over the new `$GroupScopableSubjectName` in `packages/schemas`, so a subject made scopable there fails to compile in the api until its group field is named. `all` and `Instrument` cannot be scoped; the schema refuses it and the factory throws on a stored row that has it anyway.
- Writes move to `PUT /v1/users/:id/permissions`, gated on `manage all`: an `update User` grant is one of the things this route hands out, so it must not be enough to reach it. `additionalPermissions` is removed from `$UpdateUserData`.
- A scoped grant is accepted only for a group the user belongs to; leaving a group drops the grants scoped to it.

## Web

- `/admin/users/$userId` replaces the manage sheet: an identity card, the profile form (contact, groups, status, password), a permissions card, and a delete card.
- The permissions card is a table whose last row is the composer: action, resource and scope selects sit under the headers that name them, the scope offers only the user's groups plus an explicit "All Groups", preselects the group when there is exactly one, and hides itself for the two resources that cannot take one. Every-group grants are marked with a globe and amber text.
- A new `Chip` component replaces libui's secondary badge on the identity card, which had no contrast in dark mode.

## Tests

- Unit: `$UserPermission` and `$GroupScopableSubjectName`; the factory's per-subject scoping table, unscoped grants and the throw; `UsersService.updatePermissions` and the pruning in `updateById`; the controller forwarding the ability; an api integration suite for the legacy read-back; the web editor, helpers and mutation hook.
- End to end: the four sheet tests now drive the page, plus page tests for granting, scope preselection, hidden scope, removal and the admin notice; the authorization spec proves a scoped grant reads one group's users and not another's, that an unscoped one reads both, that a foreign-group grant is refused, that a profile update cannot carry permissions, and that leaving a group drops its grants. The permissions route joins the privileged requests every non-admin role must be refused.

## Notes for review

- A `PATCH /v1/users/:id` body still carrying `additionalPermissions` is silently stripped by zod's default object mode rather than rejected; an e2e test pins that it grants nothing.
- Each permission on the wire now requires `groupId` (nullable, never optional), so an API script sending the old `{ action, subject }` shape gets a 400 rather than a silent instance-wide grant.
- `pnpm lint` and `pnpm test` (1217 tests) pass locally; the api integration suite and the e2e route generation need a socket bind that the agent sandbox blocked, so they were run outside it.


david-roper:
@joshunrau will this pr close issue #1355

joshunrau:
> @joshunrau will this pr close issue #1355

yes, good catch! thank you

